PR #33, opened 2014-11-11
Add travis support to automate test builds of packages [WIP]

jchodera:


kyleabeauchamp:
I think we can probably delete this PR for now as we've got a jenkins and a travis dependency checker that seems to be working.

Issue #86, opened 2015-01-10
Debugging dependency resolution


rmcgibbo:
For the sake of debugging the dependency resolution and "unsatisfiable package requirements" type stuff, a really easy way to speed up the debugging is just to replace all of the `build.sh` scripts with a no-op, so that the packages build instaneously, but still have the same dependency structure.

```
$ cat mock-packages
#!/bin/bash
for recipe in "$@"; do
    if [ -f "$recipe/build.sh" ]; then
        echo "mocking $recipe"
        echo "echo 'building $recipe'" > $recipe/build.sh
    fi;
done
$ chmod +x mock-packages
$ mock-package ./*
```

jchodera:
Great idea!

What about (1) adding this script to the repo, and (2) having travis run this?

rmcgibbo:
Good idea! One sec. I've been playing with this and I have some changes to file now.

Issue #82, opened 2015-01-09
Some packages (e.g. msmbuilder) are not being built by jenkins


jchodera:
Some packages---notably, `msmbuilder`---are not being built by jenkins.

No information about them appears in the console output:
https://jenkins.choderalab.org/job/conda-omnia-release-linux-vagrant/16/console

Not sure why this is.  Debugging.

rmcgibbo:
Not sure if this is related, but the msmbuilder recipe is a little weird in that it's just pulling the master branch from github. I'm going to do a proper release.

rmcgibbo:
@jchodera the issue looks to me like the dependency resolution in conda isn't working correctly. Can you try changing the jenkins configuration script to simply run `conda-build-all` twice. I think this will help, without a real loss of speed, since it'll skip files that have already been built.

rmcgibbo:
I can reproduce this on my machine now. I think I see the issue. Looking into it.

jchodera:
> Can you try changing the jenkins configuration script to simply run conda-build-all twice. I think this will help, without a real loss of speed, since it'll skip files that have already been built.

Done.  Jenkins `osx` and `linux` builds restarted.

rmcgibbo:
This is fixed now. You can also disable the part of the config where it's running the conda-build-all script twice.

jchodera:
> This is fixed now. You can also disable the part of the config where it's running the conda-build-all script twice.

Thanks!  I've just disabled this, so future jenkins builds will not try to build twice.

PR #87, opened 2015-01-10
Fix linearize

rmcgibbo:
1. Contains fix for the #82, without waiting for the upstream bugfix for conda-build that I filed.
2. Changes the order of iteration in conda-build-all, so that the inner loop is over the recipes and the outer loop is over python versions. This makes it easier to identify dependency resolution failures early
3. No more inline `build: script: python setup.py install`, because this prevents mocking out, ala #86)
4. Fix to an issue with the jype1 build. From https://binstar.org/omnia/jpype1/files, you can see that the python version wasn't included in the build string, even though it should have been.


jchodera:
Awesome. Is this ready to merge?

rmcgibbo:
Yep. I try adding the travis-ci / mock thing in a separate PR.

jchodera:
Rebuilding on jenkins using these new build script.

I presume I can eliminate the repeated calls to the build script from the jenkins configurations?

rmcgibbo:
I think so, but not 100% sure.

rmcgibbo:
There are still some issues on py33 with unsatisfiable specs.... could be related to https://github.com/conda/conda-build/issues/289.

jchodera:
Is this sorted now that the travis-ci passes?

PR #91, opened 2015-01-10
topo sort

rmcgibbo:
I wish this wasn't necessary...

### DIFF
--- a/conda-build-all
+++ b/conda-build-all
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-PYTHONS=( "27" "33" "34")
+PYTHONS=( "27" "33" "34" )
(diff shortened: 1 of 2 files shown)


rmcgibbo:
The  proof is in the pudding on jenkins. But I'm pretty sure that everything is fixed now.

jchodera:
Huzzah!

rmcgibbo:
I'm going to rewrite `conda-build-all` in python so that
 - it can do the topological sort without an external script, since i can't code that in bash anyways
 - optionally, before building a package, it checks if the file its about to build is already on binstar under a specified username (e.g. omnia), and if so, skips it.

This will also remove the need to have a separate version in powershell, since we can execute python on both platforms.

PR #92, opened 2015-01-10
New conda-build-all

rmcgibbo:
New version of conda-build-all.
  1. Incorporates the topological sort order, to build things in order such that the dependencies of each package exist once we try to build it.
  2. Using the `--check-against <BINSTAR_USER>` argument, this checks whether package file-to-be-built already exists under the given user's binstar. If so, it skips it. Packages which already exist in the `MINICONDA_ROOT/conda-bld/<PLATFORM>/` are also skipped (like before).

This should accompany some changes to the jenkins config
  1. This script should be called with `--check-against omnia` on jenkins
  2. `binstar upload` should not use `--force` anymore. Given that things are essentially working now, we should not overwrite packages on binstar. The proper way to build new versions of packages is to increment the version number or the build number in the `meta.yaml` file. This will trigger the build of a new version of the package, without overwriting the old version. It will also work correctly with `--check-against `.
  3. The line `conda install --yes --quiet sphinx` can be removed from the jenkins `conda-omnia-release-osx-2/configure`. It's superfluous.


rmcgibbo:
Also, what is the status of windows stuff? I can run the build on win-32 and win-64 and push the binaries tomorrow, but it would be "manual". Is that something you wanted to set up a jenkins machine for?

jchodera:
> This should accompany some changes to the jenkins config

Changes made prior to merge.

jchodera:
> Also, what is the status of windows stuff? I can run the build on win-32 and win-64 and push the binaries tomorrow, but it would be "manual". Is that something you wanted to set up a jenkins machine for?

I believe we had decided that Stanford folks would have to build the `win-*` binaries manually for the first release, but we wanted to eventually automate this using Jenkins.  I have so little expertise with this, though, that I'll have to delegate this to @bas-rustenburg once he returns from Amsterdam.

rmcgibbo:
Okay. I'll build them on my windows 8 VM.

PR #93, opened 2015-01-10
Delete mixtape recipe

kyleabeauchamp:
@rmcgibbo  we probably want to delete this at some point now that msmb3 is functional


rmcgibbo:
Good idea.

PR #95, opened 2015-01-10
Fix bug

rmcgibbo:


rmcgibbo:
This is the bug that was causing packages with names containing a hyphen to not be correctly detected as existing on binstar (so they were being rebuilt).

rmcgibbo:
Ready to merge.

jchodera:
Rebuilding again.

PR #89, opened 2015-01-10
Added travis-ci badge to README.md

jchodera:
This just adds a badge to the `README.md` to report on the travis-ci build status.

PR #96, opened 2015-01-10
Added Jenkins badges to README

jchodera:
This follows @rmcgibbo's suggestion from #89


jchodera:
Well, I'll need to tweak those badges.  Sorry...

PR #100, opened 2015-01-11
Updated omnia version to 1.0

kyleabeauchamp:


jchodera:
Should this recipe have an explicit build number?

kyleabeauchamp:
Probably doesn't matter.  If we change anything,  we should update the
minor version number.

rmcgibbo:
The conda. build number defaults to zero. If you want to increment, you just set it to 1.

rmcgibbo:
+1 to merge.

Issue #104, opened 2015-01-11
openmm-6.2-py27_0.tar.bz2 contains some erroneous files


rmcgibbo:
If you look in the openmm linux py27 package, there are some erroneous files.
```
$ wget https://binstar.org/omnia/openmm/6.2/download/linux-64/openmm-6.2-py27_0.tar.bz2
$ tar -tjvf openmm-6.2-py27_0.tar.bz2
```

In `lib/python2.7/site-packages/`, there's a full copy of `sphinx` and `sphinxcontrib-bibtex`. I don't think this is terrible, but it's not what we want.

rmcgibbo:
Also the case with the osx packages.

kyleabeauchamp:
OK, it's also happening in my manual builds, so it's gotta build our build script?

rmcgibbo:
Yeah. I will look into it. It makes sense how they'd get there, from being installed through the `build: requires`, but it's not obvious to me what we're doing wrong. I looked at the files in the msmbuilder  and mdtraj packages, and there's nothing extra there, even though they obviously have `build: requires` sections.

rmcgibbo:
Ohh. I know what it is.

rmcgibbo:
Well, maybe. Will fix soon.

PR #105, opened 2015-01-11
Fix issue with extra files in OpenMM

rmcgibbo:
I'm not 100% sure why this happens. It relates to wierd setuptools stuff like namespace packages. But this does get rid of the extra files from #104.

### DIFF
--- a/openmm/meta.yaml
+++ b/openmm/meta.yaml
@@ -36,4 +36,4 @@ test:
 
 about:
   home: http://openmm.org
-  license: MIT
+  license: LGPL and MIT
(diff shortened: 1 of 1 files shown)


rmcgibbo:
Actually, before merging this we should add a test line that does `python -m simtk.testInstallation`. That would be simple, and helpful.

kyleabeauchamp:
+1 for test

rmcgibbo:
done.

kyleabeauchamp:
+1 to merge.

I can try manual rebuild later tonight if Jenkins has more issues.  Headed
to dinner now.

rmcgibbo:
Kk.

PR #114, opened 2015-01-13
Add cluster-utils package

danielparton:


jchodera:
Do we really want to call it `cluster-utils`?  Aren't these torque-specific?  Or would we expand this to include tools for other kinds of batch queue systems?

rmcgibbo:
The requirements/run block is not correct: https://github.com/omnia-md/conda-recipes/blob/master/cluster-utils/meta.yaml#L18-L22. You don't need to list specific modules from the standard lib like `re` or `argparse`.

danielparton:
@jchodera Yes, I was thinking we could leave the repo as `cluster-utils` in case we expand it to include tools for other batch queue systems.

PR #126, opened 2015-01-15
OpenMM Windows + OPENMM_PLUGIN_DIR

rmcgibbo:
Changes to the OpenMM recipe, bumping the build number to 2. This should ensure that the [setting of environment variables](http://docs.openmm.org/6.2.0/userguide/application.html#installing-on-windows) like the PATH or OPENMM_PLUGIN_DIR is not required on *nix or windows when installing via conda.

To do this, https://github.com/pandegroup/openmm/pull/780 is incorporated as a patch file. There is also a really horrendous post-link.(bat|sh) that modifies `site-packages/simtk/openmm/version.py` when the conda package is being so that `__init__.py` knows where to find the openmm libraries.

I'll do the merge. I still need to do more testing on windows.

### DIFF
--- a/openmm/bld.bat
+++ b/openmm/bld.bat
@@ -1,14 +1,10 @@
-:: Only python 3, since python2 uses VS2008,
-:: which is not capable of compiling OpenMM
-"%PYTHON%" -c "import sys; assert sys.version_info[0] == 3"
-if errorlevel 1 exit 1
-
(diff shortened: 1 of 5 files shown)


rmcgibbo:
win-32 py34 build works!

jchodera:
> win-32 py34 build works!

Huzzah!

PR #128, opened 2015-01-16
Added mpi4py and pyyaml to requirements for yank.

jchodera:
I've added `mpi4py` (necessary if MPI runs are desired) and `pyyaml` (since we'll need it soon anyway).

### DIFF
--- a/yank/meta.yaml
+++ b/yank/meta.yaml
@@ -25,7 +25,6 @@ requirements:
     - docopt
     - gaff2xml
     - jinja2
-    - mpi4py
 
   run:
     - python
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
Should we add these requirements in the devtools version as well?  It's probably good to avoid getting out of sync

kyleabeauchamp:
Do we need mpi4py in the build section or just run?

jchodera:
Done.

jchodera:
Whoops, forgot to increment the build number for the `yank` recipe too.  Will do in a separate PR.

PR #134, opened 2015-01-16
ambermini static

rmcgibbo:
So this change, on my machine where I've trashed the file `libquadmath.0.dylib` from my homebrewed-gfortran, which causes the linker to statically link to libquadmath, produces a static build.

I can upload the binary to binstar.

### DIFF
--- a/ambermini/build.sh
+++ b/ambermini/build.sh
@@ -3,6 +3,9 @@
 # Required for cbio cluster for some stupid reason.
 #export LD_LIBRARY_PATH="/opt/gnu/gcc/4.8.1/lib64:/opt/gnu/gcc/4.8.1/lib:/opt/gnu/gmp/lib:/opt/gnu/mpc/lib:/opt/gnu/mpfr/lib"
 
+export CUSTOMBUILDFLAGS="-static-libgfortran /usr/local/Cellar/gcc/4.9.2_1/lib/gcc/4.9/libquadmath.a \
+	  -static-libgcc -lgfortran"
+
 export CFLAGS="-I$PREFIX/include $CFLAGS"
 export LDFLAGS="-L$PREFIX/lib $LDFLAGS"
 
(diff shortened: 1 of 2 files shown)


rmcgibbo:
But this is obviously not portable.

kyleabeauchamp:
OK.  And so the idea is to just append CUSTOMBUILDFLAGS to LDFLAGS and do manual builds for now?

rmcgibbo:
No, the ambermini makefile uses `CUSTOMBUILDFLAGS` itself. You don't have to append it to anything. You just need to run this recipe with a carefully-broken version of gfortran installed in the precise location `/usr/local/Cellar/gcc/4.9.2_1/lib/gcc/4.9/`.

kyleabeauchamp:
Got it

rmcgibbo:
Okay. I uploaded the file to https://binstar.org/omnia/ambermini/files.

jchodera:
What about installing gfortran as a conda package as an alternative?

rmcgibbo:
That sounds like a lot of work...

rmcgibbo:
Compilers are generally the hardest piece of software to compile.

kyleabeauchamp:
Sounds like a major project.  If that were easy to do, the Continuum people would have done it.

kyleabeauchamp:
Sounds good, but we probably need to IFDEF that flag so that it doesn't occur on linux:

```
gfortran: /usr/local/Cellar/gcc/4.9.2_1/lib/gcc/4.9/libquadmath.a: No such file or directory
gfortran: /usr/local/Cellar/gcc/4.9.2_1/lib/gcc/4.9/libquadmath.a: No such file or directory
make[1]: *** [dasum.o] Error 1
make[1]: Leaving directory `/home/vagrant/miniconda/conda-bld/work/ambermini-v.14.0.1/blas'
make: *** [libs] Error 2
Command failed: /bin/bash -x -e /home/vagrant/rmc/conda-recipes-ambermini/ambermini/build.sh
```

PR #136, opened 2015-01-17
Added requirement for cluster_utils to yank and repex

kyleabeauchamp:


jchodera:
Don't forget to increment build numbers.

kyleabeauchamp:
Given the number of changes that we are going to be making, I don't think we need to worry about building every update.

jchodera:
This is going to become ridiculously confusing if we don't stick to a consistent scheme of ALWAYS updating the build number or bumping the version number when we update the conda package.  Unless you don't expect to merge this PR for quite some time, bump the build number.

kyleabeauchamp:
Done

PR #138, opened 2015-01-19
Added more dependencies for itctools.

bas-rustenburg:


kyleabeauchamp:
update the build number on pint

bas-rustenburg:
Done.

behave/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+  # If this is a new build for the same version, increment the build
+  # number. If you do not include this key, it defaults to 0.
+  # number: 1
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - parse >=1.6.3
+    - parse_type >=0.3.4
```
kyleabeauchamp: what is this

kyleabeauchamp:
OK, so we have a weird dependency: enum34 is a backported version of `import enum` from py34 to py33 and py27.  Anaconda already has builds of enum34 available (http://repo.continuum.io/pkgs/free/linux-64/index.html).  However, the question is how to handle this in conda recipes that require enum34--python3.4 doesn't have this package because it's part of the standard library.  Anyone have thoughts?

rmcgibbo:
Use the magic comment `# [not py34]` after the dependency.

rmcgibbo:
similar to
```
requirements:
  # Packages required to build the package. python and numpy must be
  # listed explicitly if they are required.
  build:
    - python
  # Packages required to run the package. These are the dependencies that
  # will be installed automatically whenever the package is installed.
  run:
    - python
    - argparse # [py26]
```
from  http://conda.pydata.org/docs/build.html

bas-rustenburg:
Thanks!

kyleabeauchamp:
Looks good, built everything locally and uploaded

rmcgibbo:
> Looks good, built everything locally and uploaded

If you just merge it, isn't jenkins supposed to handle that?

kyleabeauchamp:
Yes, but I wanted to test things faster.

rmcgibbo:
Kk. LGTM.

jchodera:
Thanks, everybody!

@rmcgibbo : You OK with us adding packages that deal with purely experimental data to `omnia` for now?  It's easiest for us to do this until we figure out what to do with these tools (which are mainly for controlling robots, setting up ITC experiments, analyzing experimental data, etc.).

rmcgibbo:
Yeah, I'm good. We can decide how to "brand" the superpackage, and what projects should be highlighted in the documentation, webpage, etc, given that the front page of the webpage only has a finite amount of space above the fold. But that's a separate thing.

For the packages on binstar and the recipes, there's really no disadvantage in having more packages in the omnia binstar. Especially once we've got all the infrastructure set up once for this packaging stuff, there's no reason to split it up. I guess the only disadvantage would be if things are very difficult to build (lots of fortran code or something), but that doesn't seem to be the case.

jchodera:
> Yeah, I'm good. We can decide how to "brand" the superpackage, and what projects should be highlighted in the documentation, webpage, etc, given that the front page of the webpage only has a finite amount of space above the fold. But that's a separate thing.

I imagine we'd omit the non-simulation-related packages from all of that.

> For the packages on binstar and the recipes, there's really no disadvantage in having more packages in the omnia binstar. Especially once we've got all the infrastructure set up once for this packaging stuff, there's no reason to split it up. I guess the only disadvantage would be if things are very difficult to build (lots of fortran code or something), but that doesn't seem to be the case.

That's what we're thinking right now.  There's significant overhead for build a separate infrastructure to host this other stuff right now, and it currently doesn't distract significantly from omnia proper.

jchodera:
We good to merge, @bas-rustenburg?

bas-rustenburg:
Yes, this should be all dependencies itctools has.  I haven't added itctools itself yet, but I am still debugging that recipe. I intend to add it later.

jchodera:
Thanks!  This will get jenkins to build and test things.

PR #145, opened 2015-02-03
Add schema recipe.

bas-rustenburg:
I would like to build this package so I can use it for my bayesian ITC code.


bas-rustenburg:
Travis passes. Anyone object to merging?

bas-rustenburg:
Thanks!

PR #146, opened 2015-02-10
Added assaytools recipe

jchodera:
This PR adds a recipe for [assaytools](http://github.com/choderalab/assaytools).

### DIFF
--- a/assaytools/meta.yaml
+++ b/assaytools/meta.yaml
@@ -1,0 +1,41 @@
+package:
+  name: assaytools
+  version: !!str 0.1.0
+
+build:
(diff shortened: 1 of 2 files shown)


assaytools/meta.yaml
```diff
@@ -0,0 +1,41 @@
+package:
+  name: assaytools
+  version: !!str 0.1.0
+
```
rmcgibbo: isn't this missing a section on where to find the source?

jchodera: Whoops!  It is!

PR #147, opened 2015-02-11
Update versions of convex optimization packages

rmcgibbo:


rmcgibbo:
This version of cvxpy adds py3 support, so that cleans up a previous wart.

jchodera:
Looks good, seems to pass!

PR #148, opened 2015-02-11
Script to check for outdated packages

rmcgibbo:
This adds a script to check against PyPI if any of our recipes that pull their source from PyPI are outdated. Not that we necessarily have to update everything, but we might at least want to look at the release notes.

Current output:
```
$ ./check-outdated
behave-1.2.4 outdated. version 1.2.5 is on pypi
cvxpy-0.2.15 outdated. version 0.2.17 is on pypi
ecos-1.0.5 outdated. version 1.1.0 is on pypi
numdifftools-0.6.0 outdated. version 0.7.7 is on pypi
pybtex-docutils-0.2.0 outdated. version 0.2.1 is on pypi
python-coveralls-2.4.3 outdated. version 2.5.0 is on pypi
scs-1.0.6 outdated. version 1.0.7 is on pypi
```


jchodera:
Cool!

I wonder if there's a way to automatically rebuild all the `pypi`-derived packages with `conda skeleton`...

rmcgibbo:
I'm skeptical that it's worth it. I want to actually read the release notes before updating packages.

PR #156, opened 2015-02-20
Changed checkout method for openmm source.

jchodera:


jchodera:
This fixes a build issue on jenkins.

PR #164, opened 2015-03-02
added conda recipe for svgwrite

jhprinz:
added svgwrite to replace pip install svgwrite for `openpathsampling`

A build package has already been added to `binstar/omnia`

The recipe was built using `conda skeleton` but I had to remove the `svgwrite/data` import for testing.
The package seems to load correctly.


rmcgibbo:
LGTM

PR #165, opened 2015-03-19
Added ensembler recipe

danielparton:


danielparton:
Note: runtime requirement `subprocess32` is not available through the default conda channels. However, the current version of the `ensembler` recipe does not run tests, so this should not be a problem for now.

I'll try to think of a workaround which could be implemented in a future version of this recipe.

kyleabeauchamp:
Will merge once travis done

jchodera:
Can we add a recipe for subprocess32?

jchodera:
subprocess32 is on pypi, so you should be able to trivially create a conda
recipe using 'conda skeleton' and add that to omnia:

http://conda.pydata.org/docs/commands/build/conda-skeleton.html

kyleabeauchamp:
@danielparton  and I discussed this and thought it makes sense to postpone for now, as we're still discussing the various ways to incorporating these features.

danielparton:
Ok, I've added a recipe for `subprocess32` - waiting for Travis now.

jchodera:
Tests pass, so merging!

PR #174, opened 2015-04-21
Don't run tests for pyhmc

mpharrigan:
There are a bunch of packages that aren't
available?

ref: https://github.com/rmcgibbo/pyhmc/blob/master/devtools/conda-recipe/meta.yaml

### DIFF
--- a/pyhmc/meta.yaml
+++ b/pyhmc/meta.yaml
@@ -20,15 +20,15 @@ requirements:
     - python
     - setuptools
     - numpy
+    - scipy
 
 test:
   requires:
     - nose
     - scipy
+    - statsmodels
   imports:
     - pyhmc
-  commands:
-    - nosetests pyhmc.tests
 
 about:
   home: https://github.com/rmcgibbo/pyhmc


rmcgibbo:
I'm not sure I understand your question.

mpharrigan:
```
    - rpy2
    - r-coda        # spectrum0, spectrum0.ar, effectiveSize
    - r-mcmc        # initseq()
    - r-batchmeans  # bm
```

don't seem to be available, and pyhmc wouldn't build because it couldn't run the tests.

rmcgibbo:
Looks like it requires `onda config --add channels r` https://github.com/rmcgibbo/pyhmc/blob/master/.travis.yml

kyleabeauchamp:
Is this good to go?

mpharrigan:
@kyleabeauchamp I think we need to decide whether disabling the tests here is appropriate

@rmcgibbo the travis.xml looks pretty involved for getting the tests to run. Do you want to disable the tests here and relegate them to travis only?

rmcgibbo:
Okay, fair enough.

PR #175, opened 2015-04-21
pyvisfile and dependencies

rmcgibbo:
I want to use http://mathema.tician.de/software/pyvisfile/ in a project I'm working on. Unfortunately it's a little tricky to compile all the dependencies, so I made conda packages so I don't have to do it again.

I'm not sure exactly if we've decided on the scope of this repo, but this seemed like a better place to put the recipes than anywhere else?

### DIFF
--- a/appdirs/bld.bat
+++ b/appdirs/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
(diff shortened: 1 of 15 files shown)


jchodera:
Sounds reasonable!  We have a few packages like that in here too.

rmcgibbo:
I'll test the build out on Mac before merging.

jchodera:
This looks good to me.

rmcgibbo:
Okay. Tested fine on os-x py27 and linux-64 py34, so I think this is good to merge.

PR #176, opened 2015-04-22
Changed OpenMM recipe to build osx version with 10.7 SDK instead of 10.9

jchodera:
This addresses https://github.com/pandegroup/openmm/issues/881

### DIFF
--- a/openmm/build.sh
+++ b/openmm/build.sh
@@ -14,8 +14,8 @@ if [[ "$OSTYPE" == "linux-gnu" ]]; then
     CMAKE_FLAGS+=" -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-6.5"
 elif [[ "$OSTYPE" == "darwin"* ]]; then
     CMAKE_FLAGS+=" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++"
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
+1

jchodera:
Any idea why this is erroring?

jchodera:
It's timing out trying to download packmol:
https://travis-ci.org/omnia-md/conda-recipes/builds/59592683#L1545-L1560

Can we put a local cache of this somewhere?

kyleabeauchamp:
Is this the first time we've had a problem with packmol?  If it's the first time, I'd say try to rebuild the travis rather than doing more customization of our CI tools.  

If this has been causing problems repeatedly, then we can follow the model here: https://github.com/mdtraj/mdtraj/blob/master/devtools/travis-ci/install.sh#L27

jchodera:
Haven't seen this before. I've triggered a travis rebuild.

rmcgibbo:
You should increment the build number in the `meta.yaml` too, I think.

jchodera:
Good catch!

jchodera:
Seems to pass.  Committing.

jchodera:
We should change the `openmm-dev` recipe too.  On it.

PR #181, opened 2015-04-23
[WIP] Incremental binstar uploads

jchodera:
I'm investigating how easy it would be to allow incremental binstar uploads by adding an additional `--upload` flag.

Still a work in progress.


rmcgibbo:
I did this already in one of my appveyor related commits. It has the `--upload` flag now, and optionally pulls the token from the environment variable `BINSTAR_TOKEN`. Originally I had been passing the token on the command line, but since the commands are echoed to the log during the build, that "publically" exposes the secret token.

jchodera:
Awesome!

PR #191, opened 2015-05-08
binstar upload

rmcgibbo:
Add a flag to `conda-build-all` to enable upload direct to binstar upon completion of each build:

usage: `$ python conda-build-all --upload omnia ...`


rmcgibbo:
Actually to work with our stuff this needs another argument for the token.

rmcgibbo:
messed up git history :(

PR #199, opened 2015-05-10
Appveyor: fix install.ps1

rmcgibbo:


mpharrigan:
out of curiousity, why was this broken?

rmcgibbo:
the `+` doesn't actually concatenate strings in powershell.

Issue #197, opened 2015-05-09
Add "skip list"


rmcgibbo:
We should add a "skip list" somewhere of packages that either don't build on windows, don't build on py3, or whatever.

As it is currently, on every commit all of the CI systems waste a lot of effort downloading the prerequisites and setting up the build only to have it fail for some of these combinations that we know aren't supported.

jchodera:
What about having each package recipe have a file that lists supported versions of things, like Python versions and platforms?  Like
```
osx-64 win-32 linux-64
py27 py33 py34
```

rmcgibbo:
I don't really want to mess with the conda recipe format though. Also, the assumption that "unless otherwise specified, try to build it with every combination" seems better than the alternative.

jchodera:
Why would this mess with the recipe format?  Wouldn't additional files be ignored?

We can still have the script try to build everything if this file is not present in the recipe directory.

I think we should avoid having one big file that aggregates information about all recipes.  Having separate files in each recipe directory ensures that individual package maintainers can control what gets built for their package (if desired) without fear of collisions.

rmcgibbo:
Closed by #200

PR #200, opened 2015-05-10
ENH: Add skiplist feature

rmcgibbo:
Add feature to `conda-build-all` to implement #197. Following @jchodera's suggestion, there's no single central list. Each recipe directory is now permitted to contain a file called "skip.yaml" which can contain a list of platforms that are to be skipped.

Each skip specification is supposed look like `<PLATFORM_SPEC>_<PYTHON_SPEC>`, `<PLATFORM_SPEC>`, or `<PYTHON_SPEC>`, e.g. `win-32_py34`, `win-32`, or `py34`.

I also added some `skip.yaml` files to particular packages that don't support windows, python2, python3 or some particular subset, including

 - ambermini (no windows)
 - openmm (no windows on py2)
 - pyemma (no py3)
 - pytram (no py3)
 - pymbar (no py3)
 - sh (no windows)
 - subprocess32 (no windows or py3)


jchodera:
This is awesome, but wouldn't it make more sense to have the lists contain which platforms are to be *built*?  Otherwise, all skiplists will have to be updated if some new platform (e.g. `py35`) is introduced.

Lack of a list would of course still mean that all combinations are attempted, as usual.

rmcgibbo:
I don't like that as much. When a new platform comes out, why not take the presumption that we'll give it a try and see if things work? I like the idea that in general, we try all combinations. This is just about carving out limited exceptions to that.

jchodera:
Sure, let's give that a try.

peastman:
I notice you're having it skip all Windows builds for OpenMM.  What's the difficulty with building on Windows?  It would be really nice if we could get that working.

rmcgibbo:
No, just skipping python2 on windows, since we don't support VS2008. https://github.com/omnia-md/conda-recipes/blob/master/openmm/skip.yaml

peastman:
Ok, I missed that.  Thanks.

Issue #203, opened 2015-05-11
Why do we have openmm-dev?


rmcgibbo:
Why do we have the openmm-dev package here? It's the only "-dev" package that we have in this repo -- seems a little out of place.

Given that
 (a) this repo isn't triggered on commits to pandegroup/opennmm and;
 (b) the build scripts in this repo only rebuild a package when it isn't on binstar -- meaning that you need to bump the version or build number to get a new build activated;

it seems like whatever the intention of having an openmm-dev recipe in this repo, it doesn't make sense.

jchodera:
See https://github.com/pandegroup/openmm/pull/817

rmcgibbo:
I think my two points still apply. The recipe isn't doing what you want it to.

rmcgibbo:
https://jenkins.choderalab.org/job/conda-openmm-dev-linux-vagrant looks like it's solving the root problem here, and is totally unrelated to this repo.

jchodera:
`openmm-dev` is being built on Jenkins at the request of @peastman.

(a) building of `openmm-dev` is [triggered by commits to the openmm github repo](https://jenkins.choderalab.org/job/conda-openmm-dev-linux-vagrant/configure)

(b) the fact that the omnia build Jenkins job doesn't build `openmm-dev` is irrelevant because it gets built in a separate Jenkins job

We just dumped `openmm-dev` in the `conda-recipes` github repo because we needed someplace with version control that I had write access to get it working.

jchodera:
I'd honestly rather have `openmm-dev` built by travis and appveyor on commits to the openmm repo instead of the Jenkins hack.

jchodera:
I do agree that it's out of place in [omnia/conda-recipes](https://github.com/omnia-md/conda-recipes), but it's less out of place there than anywhere else it can reasonably go write now.

rmcgibbo:
Out of the 55 recipes in this repository, 54/55 of them are designed to work properly with the two omnia-release jenkins hooks and appveyor hook referenced in the README, and 1/55 doesn't fit the model at all.

It's pretty easy to create a new repo.

jchodera:
OK, will move to new repo.  

How about `omnia/conda-dev-recipes`?

rmcgibbo:
Or just openmm-dev?

jchodera:
We don't really want to have a separate repo for every special case we need built via Jenkins, do we?

rmcgibbo:
Fixed by #204.

PR #204, opened 2015-05-11
Removed openmm-dev package

jchodera:
This is an attempt to fix https://github.com/pandegroup/openmm/issues/910


jchodera:
The openmm-dev package has now been removed due to #203

jchodera:
It is now here: https://github.com/omnia-md/openmm-dev

PR #205, opened 2015-05-12
Fix OpenMM post-link.{sh,bat} files

rmcgibbo:


jchodera:
You forgot to back-port those fixes into the [openmm-dev](https://github.com/omnia-md/openmm-dev) recipe.

PR #210, opened 2015-06-02
Lowercase parmed

kyleabeauchamp:


rmcgibbo:
+1 to merge and try it out, or should we wait for @swails?

kyleabeauchamp:
I think I'm just going to try it--he can't really debug this stuff himself as he doesn't have full access to jenkins.

rmcgibbo:
Sounds good. (But for reference, you can debug it locally by just running conda build; there shouldn't be anything _too_ special about the jenkins server, except for the binstar token)

PR #211, opened 2015-06-02
Fix version string for parmed

kyleabeauchamp:


kyleabeauchamp:
Can't have dash in version, causes fatal error when I do `conda-build parmed`

PR #213, opened 2015-06-04
Increment bhmm to 0.3.0; fix bhmm weirdness

jchodera:
Still debugging this.


jchodera:
Is it OK that `appveyor` failed?  It looks like the issue is that `openmmtools` was not found in the `win-32` channel.

rmcgibbo:
yeah, that's fine. the appveyor thing still isn't 100%.

PR #220, opened 2015-06-10
Dont run tests on pymbar conda

kyleabeauchamp:


jchodera:
We can keep the tests for the `pymbar-dev` conda package.

PR #222, opened 2015-06-15
Upgrade ParmEd conda package to beta1.

swails:


jchodera:
Just noticed this!  Sorry for not merging earlier.  We'll get you merge access.

swails:
Thanks! Btw, the turnaround time on this was not even close to too long.

PR #225, opened 2015-06-17
Update openmoltools for 0.6.8 point release

davidlmobley:
This point release is needed for ParmEd 2.0 compatibility, among other things.

### DIFF
--- a/openmoltools/meta.yaml
+++ b/openmoltools/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: openmoltools
-  version: 0.6.7
+  version: 0.6.8
 
 source:
     git_url: https://github.com/choderalab/openmoltools.git
-    git_tag: v0.6.7
+    git_tag: v0.6.8
 
 
 build:


davidlmobley:
@kyleabeauchamp - what's necessary to get the conda version to actually update? i.e. I'm not seeing any change yet on binstar (https://binstar.org/omnia/openmoltools/files).

jchodera:
Let me take a look at this.

Are you hoping for the updated openmoltools or openmoltools-dev?

davidlmobley:
Updated openmoltools. Made 0.6.8 release, updated conda recipe, but nothing
has showed up on omnia. I don't quite know what steps happen (behind the
scenes?) between updating the conda recipe and getting a new
conda-installable version, so perhaps there's something else I need to do?

Thanks.


On Wed, Jun 17, 2015 at 8:43 AM, John Chodera <notifications@github.com>
wrote:

> Let me take a look at this.
>
> Are you hoping for the updated openmoltools or openmoltools-dev?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/omnia-md/conda-recipes/pull/225#issuecomment-112855349>
> .
>



-- 
David Mobley
dmobley@gmail.com
949-385-2436

jchodera:
Jenkins must be stuck.

I'll take a look and then summarize the steps you can use yourself in the
future to at least diagnose what is going on.

jchodera:
@davidlmobley : On the front page of the [omnia-md/conda-recipes](https://github.com/omnia-md/conda-recipes) repo, there are badges in the `README.md` section for Jenkins `osx` and `linux` builds.

If you click on the badge, you get to the latest console, where you can scroll or search for `openmoltools`:
https://jenkins.choderalab.org/job/conda-omnia-release-linux-vagrant/125/console

You'll see errors like this:
```
===== testing package: openmoltools-0.6.8-py34_0 =====
import: u'openmoltools'
  File "/home/vagrant/miniconda/conda-bld/test-tmp_dir/run_test.py", line 25, in <module>
    import openmoltools
  File "/home/vagrant/miniconda/envs/_test/lib/python3.4/site-packages/openmoltools-0.6.8-py3.4.egg/openmoltools/__init__.py", line 28, in <module>
    from openmoltools import amber_parser, system_checker, utils, packmol, openeye, amber, cirpy, gromacs
  File "/home/vagrant/miniconda/envs/_test/lib/python3.4/site-packages/openmoltools-0.6.8-py3.4.egg/openmoltools/gromacs.py", line 9, in <module>
    import parmed
ImportError: No module named 'parmed'
```
Did you not fully update everything to install and import `parmed`?

Or is it possible that the new `parmed` release hadn't been built yet?  If so, I've triggered a rebuild.

davidlmobley:
Ah, submitted a fix. Looks like the omnia conda recipe hadn't been updated when `parmed` was added as a requirement. (I didn't add `parmed` as a requirement - this happened earlier - so I hadn't looked at whether it was listed in this recipe. It's unclear to me though why this didn't result in failures earlier; probably it's because until now it was only used by one minor function in `openmoltools`.)

davidlmobley:
(Not sure if you want me to be able to, but I can't currently merge PRs on here so you'll have to do so.)

davidlmobley:
Now I see why it wasn't caught before. Previously the only `import parmed` was buried in a function:

```
def randomize_mol2_residue_names(mol2_filenames):
    """Find unique residue names for a list of MOL2 files.  Then
    re-write the MOL2 files using ParmEd with the unique identifiers.
    """
    import parmed
```

so presumably this wasn't triggered on building. Ugh. And since the package's conda recipe was correct, Travis-CI wasn't catching the problem. 

Is there a good reason for the `meta.yaml` in `openmoltools` to be distinct from the one in the omnia conda-recipes? i.e. should they normally be exact duplicates? It seems like where things went wrong here is that they are not duplicates.

kyleabeauchamp:
It's good to keep them as synced as possible.  In some situations, there are reasons for brief divergences , but otherwise one should keep them synced.

jchodera:
The reason they are separate, as @kyleabeauchamp, are primarily twofold:
* The package name is different (`packagename-dev` vs `packagename`) to distinguish the most-up-to-date packages from official releases.  We tried a bunch of variations on version numbering schemes for just `packagename`, but conda has a finnicky system for determining which package you get if you say `conda install packagename` that meant that none of our other attempts to have release versions and dev versions, with release versions always delivered to you if you don't specify a version number, were viable.
* The recipe in the repo's `devtools/conda-package/` might require bleeding-edge features from other packages (e.g. `openmm-dev` instead of `openmm`).

Other than this, you should make sure the recipes are *identical* with the `omnia-md/conda-recipes` recipes.

PR #229, opened 2015-06-18
Require development openmm (6.3) which ParmEd needs

davidlmobley:

PR #230, opened 2015-06-18
Added skip.yaml to msmbuilder recipe

jchodera:
Noted that `win64` wasn't building in #229, so added a `skip.yaml`.


mpharrigan:
Let's remember to remove this when we release the next version of msmbuilder

jchodera:
I can't seem to figure out why appveyor errored out here.

rmcgibbo:
i presume it's a timeout.

PR #231, opened 2015-06-18
updated pyemma and pytram. Removed pystallone

marscher:


marscher:
actually nothing really failed, despite hitting the maximum execution time during openmm building.

jchodera:
OK, Jenkins should build now!

PR #233, opened 2015-06-19
Find the appropriate version of conda_build

rmcgibbo:
Modifies `conda-build-all` to find conda_build not by taking the first thing in the path, but by looking adjacent to the python interpreter used to invoke the script. The reason for this is, on windows, that I have both a 64-bit and 32-bit conda installed. Only the 64-bit interpreter is in my `PATH`. This way, I can do `C:\Path\To\32\bit\python.exe conda-build-all <packages>`, and it will invoke the correct copy of conda_build (the 32-bit one), even though it's not in my path.

### DIFF
--- a/conda-build-all
+++ b/conda-build-all
@@ -13,6 +13,7 @@ import tempfile
 from subprocess import check_output, call, check_call, CalledProcessError
 from functools import reduce
 from six import string_types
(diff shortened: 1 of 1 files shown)


jchodera:
Cool!

PR #237, opened 2015-06-27
Ensembler v1.0.1 release

danielparton:
Removes mpi4py as a dependency. This should fix issues with Jenkins OS X build.

### DIFF
--- a/ensembler/meta.yaml
+++ b/ensembler/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: ensembler
-  version: !!str 1.0
+  version: !!str 1.0.1
 
 # TODO update
 source:
    git_url: https://github.com/choderalab/ensembler
-   git_tag: v1.0
+   git_tag: v1.0.1
 
 build:
   entry_points:
@@ -18,13 +18,12 @@ requirements:
 
   run:
     - python
-    - mpi4py
     - mdtraj
     - msmbuilder
     - biopython
     - openmm
     - pdbfixer
-    - numpy
+    # - numpy
     - lxml
     - pyyaml
     - docopt


jchodera:
I don't think `mpi4py` was the problem, was it?  There was some HTTP communication error grabbing the `msmbuilder` package from `binstar`, which I don't think was related to `mpi4py`.  `yank` builds fine with `mpi4py` as a dependency, I believe.

jchodera:
Whoop, you were right: There was an error message regarding MPI:
```
--------------------------------------------------------------------------
Sorry!  You were supposed to get help about:
    opal_init:startup:internal-failure
But I couldn't open the help file:
    /opt/anaconda1anaconda2anaconda3/share/openmpi/help-opal-runtime.txt: No such file or directory.  Sorry!
--------------------------------------------------------------------------
[mski1620:36657] [[INVALID],INVALID] ORTE_ERROR_LOG: Error in file runtime/orte_init.c at line 79
--------------------------------------------------------------------------
Sorry!  You were supposed to get help about:
    mpi_init:startup:internal-failure
But I couldn't open the help file:
    /opt/anaconda1anaconda2anaconda3/share/openmpi/help-mpi-runtime: No such file or directory.  Sorry!
--------------------------------------------------------------------------
```
I ran into this recently too.  Apparently, `from mpi4py import MPI` doesn't gracefully throw an exception anymore when this fails.  In `yank`, we just decided to work around this by having the use specify an `--mpi` flag, and checking whether this flag was specified to see if we should import MPI.

danielparton:
My approach was to write a dummy MPI class which is used if `import mpi4py.MPI` fails. This avoids the need to encapsulate all MPI-enabled functionality with `if mpi:` statements. If mpi4py fails to load, then anything with MPI functionality will basically act as if there is a single MPI process.

```.py
class MPIState:
    def __init__(self):
        import mpi4py.MPI
        self.comm = mpi4py.MPI.COMM_WORLD
        self.rank = self.comm.rank
        self.size = self.comm.size

class DummyMPIState:
    def __init__(self):
        self.comm = DummyMPIComm()
        self.rank = 0
        self.size = 1

class DummyMPIComm:
    def Barrier(self):
        pass

    def bcast(self, obj, root=0):
        return obj

    def gather(self, obj, root=0):
        return [obj]

try:
    mpistate = MPIState()
except Exception as e:
    logger.debug('WARNING: failed to initialize MPIState:\n{0}'.format(e))
    mpistate = DummyMPIState()
```

jchodera:
OK, so you are dependent on `import mpi4py.MPI` failing with an exception if MPI is not available.

I guess that means the code will not work on `osx` until the `mpi4py` issue is fixed?

I've reported this here, though I'm not sure if this is the proper place to do so:
https://github.com/Anaconda-Server/docs.anaconda.org/issues/85

Oddly enough, the binstar packages haven't been updated in six months:
https://binstar.org/anaconda/mpi4py/files

Perhaps the issue is with the `openmpi` package instead?  but this was updated six months ago too...
https://binstar.org/anaconda/openmpi/files

danielparton:
Oh, you're right - `import mpi4py.MPI` doesn't fail with an exception in this case.

There's an issue about this here:
https://github.com/ContinuumIO/anaconda-issues/issues/96

danielparton:
Two options I can think of:
1. Write a function which checks whether Ensembler is running on OS X, and then checks whether the directory /opt/anaconda1anaconda2anaconda3 exists. If it doesn't exist, the dummy MPI class is used (and no attempt is made to import mpi4py.MPI).
2. Implement the `--mpi` flag thing, with `if` statements on all MPI-enabled code.

danielparton:
So the problem here is that if a user is running OS X with Anaconda and mpi4py, then upon import mpi4py will look for the directory `/opt/anaconda1anaconda2anaconda3`. A workaround is to create a symlink from that path to the user's Anaconda installation.

I've written some updated code (not yet checked in) which should handle this more effectively.

To summarize the most important cases (all on OS X using Anaconda):
* If a user has not installed mpi4py at all, then DummyMPIState is used.
* If a user has installed mpi4py and the `/opt/anaconda1anaconda2anaconda3` directory is available, then the MPIState is used - this attempts to import mpi4py.MPI, and should be successful
* If a user has installed mpi4py but the above directory does not exist, then a warning is printed (including instructions to create a symlink), and Ensembler continues to run while using the DummyMPIState class. This occurs whether using `mpirun` or not.
* If a user tries to use the `mpirun` program from the conda `openmpi` package and has not created the above symlink, then `mpirun` will abort (before even running Ensembler) with the error message @jchodera pasted above. This is a problem with the conda mpi4py package, not Ensembler, so there's not much I can do about this directly. But I've added a warning in the Ensembler documentation to suggest creating the symlink if a user comes across this problem.

Does this sound ok?

```.py
class MPIState:
    def __init__(self):
        import mpi4py.MPI
        self.comm = mpi4py.MPI.COMM_WORLD
        self.rank = self.comm.rank
        self.size = self.comm.size

class DummyMPIState:
    def __init__(self):
        self.comm = DummyMPIComm()
        self.rank = 0
        self.size = 1

class DummyMPIComm:
    def Barrier(self):
        pass

    def bcast(self, obj, root=0):
        return obj

    def gather(self, obj, root=0):
        return [obj]

try:
    import imp
    imp.find_module('mpi4py')   # just looks for module, does not try to import it. Raises ImportError if module not found
    import platform
    if (
            platform.system() == 'Darwin' and
            not os.path.exists(os.path.join(os.path.sep, 'opt', 'anaconda1anaconda2anaconda3'))
        ):
        logger.warning(
            'WARNING: ignoring mpi4py module.\n'
            'Using mpi4py on OS X with Anaconda currently requires that'
            '/opt/anaconda1anaconda2anaconda3 points to your Anaconda installation.\n'
            'As a workaround, you can create a symlink, e.g. '
            '"sudo ln -s ~/anaconda /opt/anaconda1anaconda2anaconda3"'
        )
        mpistate = DummyMPIState()
    else:
        mpistate = MPIState()
except ImportError:
    mpistate = DummyMPIState()
```

danielparton:
(Don't merge this yet, as I'll need to create a new Ensembler 1.0.1 release first.)

jchodera:
I think this solution is fine for now.  Thanks!

danielparton:
Alright, I've committed that code into Ensembler and made a new `v1.0.1` release. Should be ok to merge this PR now.
Thanks for checking this over! This mpi4py bug is annoying...

jchodera:
Agreed!

danielparton:
Oh man. So apparently upon running `conda install`, any instances of `/opt/anaconda1anaconda2anaconda3` present in strings within the package will be replaced by the user's Anaconda install directory.

http://conda.pydata.org/docs/building/spec.html
(see the `info/has_prefix` section)

So when I installed the new Ensembler release in a fresh conda environment and tested out the mpi4py warning, I got the following message. The above directory path has been replaced with the path to my new conda environment.

```
>>> import ensembler
WARNING: ignoring mpi4py module.
Using mpi4py on OS X with Anaconda currently requires that/Users/partond/anaconda/envs/ensemblertest points to your Anaconda installation.
As a workaround, you can create a symlink, e.g. "sudo ln -s ~/anaconda /Users/partond/anaconda/envs/ensemblertest"
```

(There was also a space missing after "that")

I've made a workaround by specifying the path with os.path.join. See below.
I've committed this and made a new 1.0.1 release.

@jchodera Could you schedule a new OS X build on Jenkins please? (I think that is the right thing to do?)

Sorry about this. I didn't notice it before, since this problem doesn't occur when installing from source with `python setup.py install`

```.py
try:
    import imp
    imp.find_module('mpi4py')
    import platform
    anaconda_placeholder_path = os.path.join(os.path.sep, 'opt', 'anaconda1anaconda2anaconda3')
    if platform.system() == 'Darwin' and not os.path.exists(anaconda_placeholder_path):
        logger.warning(
            'WARNING: {0} not found.\n'
            'Ignoring mpi4py.\n'
            'Using mpi4py on OS X with Anaconda currently requires that '
            '{0} points to your Anaconda installation.\n'
            'As a workaround, you can create a symlink, e.g. '
            '"sudo ln -s ~/anaconda {0}'.format(
                anaconda_placeholder_path
            )
        )
        mpistate = DummyMPIState()
    else:
        mpistate = MPIState()
except ImportError:
    mpistate = DummyMPIState()
```

danielparton:
pinging @jchodera - I had to update the Ensembler 1.0.1 release again. So I think next step would be to schedule a new omnia build on Jenkins, and I think only you have permissions to do that? At least, I don't see a way to do that when I look at the Jenkins webpages.

jchodera:
@danielparton : If you want to replace the 1.0.1 package, you *must* update the ensembler conda recipe and increment the build number.  Open a PR to do that?

PR #244, opened 2015-07-06
[pymbar] Renamed pymbar -> pymbar3; restored pymbar2

jchodera:
The `pymbar2` packages disappeared off binstar, so for compatibility, we need to retain both `pymbar2` and `pymbar3` recipes right now.


jchodera:
Something funny is going on with travis-ci, but it's not entirely clear that it is related to this PR.

jchodera:
I don't understand why this is still failing...

rmcgibbo:
The travis checks for this repo are just worthless. We should get rid of them.

Issue #245, opened 2015-07-06
Ambermini recipe md5 mismatch


jchodera:
I just realized (from looking at travis-ci errors) the updated `ambermini` recipe has an md5 mismatch:
```
Downloading https://github.com/choderalab/ambermini/archive/v.14.0.1.zip
Success
Error: MD5 mismatch: '527128edf03becedda26016a366cb9f1' != '4653df08b0f9d94a39704c10591fcd84'
```

jchodera:
Solved.

PR #248, opened 2015-07-07
[ambermini] Updated to 15.0.0

jchodera:
This updates `ambermini` to 15.0.0 to match the AmberTools 15 release.


jchodera:
Also addresses #245

PR #252, opened 2015-07-09
[bhmm] Update to 0.4.0

jchodera:
We've recently split the BHMM part into its own toolkit (`bhmm`).


jchodera:
Don't merge just yet.

PR #253, opened 2015-07-10
[ambermini] Update to 15.0.3

jchodera:


jchodera:
I think something weird is going on with travis building this PR.  AppVeyor passed and there doesn't appear to be a broken dependency, so merging.

rmcgibbo:
The travis thing is more trouble than it's worth, and I don't think appveyor is actually working. It's just trying and failing to build OpenMM over and over I think.

Does ambermini require a fortran compiler?

jchodera:
> Does ambermini require a fortran compiler?

Yes.  `gfortran`.  We have that set up on Jenkins, but I realize that this might require something to be installed for Appveyor.  Apologies for the lack of warning.

jchodera:
> I don't think appveyor is actually working. It's just trying and failing to build OpenMM over and over I think.

If we build an OpenMM release manually just once and upload it to binstar, that would cause the appveyor script to skip trying to build OpenMM again, right?

rmcgibbo:
Yeah. I can run through the build on my windows machine next week.

PR #255, opened 2015-07-13
Update msmtools

marscher:
also enabled container based infrastructure on travis by adding ``` sudo: false``` to travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: c
+sudo: false
 
 install:
     - MINICONDA=Miniconda-latest-Linux-x86_64.sh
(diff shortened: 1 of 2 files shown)


jchodera:
Is there a chance the container thing could break Travis?  Or do we not use
sudo at all?

marscher:
Am 13.07.2015 um 18:02 schrieb John Chodera:
> Is there a chance the container thing could break Travis? Or do we not use
> sudo at all?
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/omnia-md/conda-recipes/pull/255#issuecomment-120980122>.
> 
not using sudo is mandatory for using the docker based infrastructure. I
do not think there is anything which can break, if one does not use sudo.

jchodera:
Looks like it worked, thanks!

PR #256, opened 2015-07-14
bump version of msmtools

marscher:


marscher:
that was fast, thank you :+1:

PR #257, opened 2015-07-15
Windows fftw3f

rmcgibbo:
I rebuilt fftw3f on windows to make static libraries, so that you don't have to modify the PATH to use them.


rmcgibbo:
Will push the built win32 binaries manually.

PR #258, opened 2015-07-16
[msmtools] build for py3.3 and py3.4

marscher:


marscher:
appveyor builds are broken for an unknown reason...

PR #260, opened 2015-07-16
Turn off Travis-CI

rmcgibbo:


jchodera:
OK, I guess this is the logical thing to do right now.

jchodera:
I guess we still have to do some additional cleanup here to turn off travis hooks and remove the travis-ci badge.

Issue #262, opened 2015-07-19
Weird pymbar3 problems


jchodera:
I am getting some strange `pymbar3` build issues.

On `osx`:
```
Installed /Users/choderaj/anaconda/envs/_build/lib/python2.7/site-packages/pymbar-3.0.0.dev0-py2.7-macosx-10.5-x86_64.egg
Processing dependencies for pymbar==3.0.0.dev0
Finished processing dependencies for pymbar==3.0.0.dev0
number of files: 54
Fixing permissions
Fixing linking of /usr/lib/libSystem.B.dylib in /Users/choderaj/anaconda/envs/_build/lib/python2.7/site-packages/pymbar-3.0.0.dev0-py2.7-macosx-10.5-x86_64.egg/_pymbar.so
Fixing linking of /opt/local/lib/libgcc/libgcc_s.1.dylib in /Users/choderaj/anaconda/envs/_build/lib/python2.7/site-packages/pymbar-3.0.0.dev0-py2.7-macosx-10.5-x86_64.egg/_pymbar.so
install_name_tool -id @rpath/python2.7/site-packages/pymbar-3.0.0.dev0-py2.7-macosx-10.5-x86_64.egg/libSystem.B.dylib /Users/choderaj/anaconda/envs/_build/lib/python2.7/site-packages/pymbar-3.0.0.dev0-py2.7-macosx-10.5-x86_64.egg/_pymbar.so

install_name_tool -add_rpath @loader_path/../../../ /Users/choderaj/anaconda/envs/_build/lib/python2.7/site-packages/pymbar-3.0.0.dev0-py2.7-macosx-10.5-x86_64.egg/_pymbar.so
error: install_name_tool: changing install names or rpaths can't be redone for: /Users/choderaj/anaconda/envs/_build/lib/python2.7/site-packages/pymbar-3.0.0.dev0-py2.7-macosx-10.5-x86_64.egg/_pymbar.so (for architecture x86_64) because larger updated load commands do not fit (the program must be relinked, and you may need to use -headerpad or -headerpad_max_install_names)

Error: install_name_tool failed with exit status 1
```
On `linux`, there is another weird Jenkins issue:
```
Traceback (most recent call last):
  File "./conda-build-all", line 307, in <module>
    main()
  File "./conda-build-all", line 88, in main
    execute(args, p)
  File "./conda-build-all", line 281, in execute
    metadata = MetaData(recipe)
[...]

PR #264, opened 2015-07-19
Update conda-build-all for multiple recipes for same package

jchodera:
This PR updates `conda-build-all` to handle cases where multiple recipe directories might encode different versions of the same package (like pymbar2, pymbar3).


jchodera:
This hopefully fixes #262 which was created by #244

jchodera:
I think this is working; the only failure seems to be lack of authentication (which does not happen in PRs).

PR #268, opened 2015-07-25
[openmm] OpenMM 6.3

rmcgibbo:
I would prefer to get this right the first time, so no merge yet.


swails:
+1.  When will these be posted?  After this gets merged, or shortly thereafter?

rmcgibbo:
This works for Linux and OSX, so it can probably be merged. Still needs some work for Win-64 I think, but we can do that later.

jchodera:
Thanks!  I'm still debugging some issues with Jenkins failing all of a sudden, so give me a few more min before we merge.

jchodera:
See #270

jchodera:
OK, I think we're sorted now.

jchodera:
Merging #269 first just to double-check everything works as intended.

jchodera:
Any chance you could retrigger the appveyor build for this PR?  I wonder if #271 fixed your issues.

PR #269, opened 2015-07-26
[pytram] release 0.2.6

cwehmeyer:


jchodera:
Stay tuned---we're fixing some issues with appveyor and Jenkins.

Issue #270, opened 2015-07-26
Jenkins issues due to binstar change to conda-server?


jchodera:
Jenkins is failing:
* On [`osx`](https://jenkins.choderalab.org/job/conda-omnia-release-osx-2/164/console):
```
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!
!!  The 'binstar' conda package has been renamed to 'conda-server'.
!!  Please run:
!!
!!    conda install conda-server
!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
* on [`linux`](https://jenkins.choderalab.org/job/conda-omnia-release-linux-vagrant/157/console):
```
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!
!!  The 'binstar' conda package has been renamed to 'conda-server'.
!!  Please run:
!!
!!    conda install conda-server
!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

I've fixed a minor issue with failing to import `six`, but now it seems that we are being forced to change `binstar` to `conda-server`

jchodera:
Shit, now they're just fucking with us:
```
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!
!!  The 'binstar' conda package has been renamed to 'anaconda-client'.
!!  Please run:
!!
!!    conda install anaconda-client
!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

mpharrigan:
oh dear

rmcgibbo:
done.

PR #271, opened 2015-07-26
Update appveyor script for binstar -> conda-build and missing 'six' problems

jchodera:
This should fix some appveyor issues.


jchodera:
This should address some of the #270 issues on the appveyor side.

PR #276, opened 2015-07-27
added pymol (which needs glew)

marscher:


jchodera:
If this won't work for `osx`, maybe we can add `skip.yaml` with the `- osx` exclusion?

If it does work for `osx`, wouldn't it require adding a conda recipe that supplies X11 packages?

marscher:
Am 27.07.2015 um 19:36 schrieb John Chodera:
> If this won't work for |osx|, maybe we can add |skip.yaml| with the |-
> osx| exclusion?
I do not know, if skip.yaml can also be used to skip certain platforms.
> 
> If it does work for |osx|, wouldn't it require adding a conda recipe
> that supplies X11 packages?
glew should compile under osx, since it is only an extension library on
top of OpenGL.

jchodera:
> I do not know, if skip.yaml can also be used to skip certain platforms.

Yes: https://github.com/omnia-md/conda-recipes/blob/master/ambermini/skip.yaml

> glew should compile under osx, since it is only an extension library on
top of OpenGL.

Let me test!

jchodera:
Looks like `glew` builds fine on `osx` but we are missing `freeglut`:
```
> conda build pymol
Removing old build directory
Removing old work directory
BUILD START: pymol-1.7.4.0-np19py27_0
Fetching package metadata: ........
Error: No packages found in current osx-64 channels matching: freeglut
```
Any chance we could add `freeglut` too?

marscher:
sorry I've forgotten that one. Wait a sec.

marscher:
mhm, freeglut is actually the one that depends on X11. So there is at
least no easy way to provide it for osx?

jchodera:
Huh: On my brand-new iMac 5K, there *are* X11 libraries installed by default, in /usr/X11 and /usr/X11R6!

jchodera:
Is there a `freeglut` conda recipe somewhere we could tweak?

marscher:
Do you also have the X11 headers?

marscher:
But it is cmake based, so it should be easy to create one.

jchodera:
Yep, we have both `/usr/X11/include/X11/` and `/usr/X11R6/include/X11/`.

peastman:
What happens if you launch a program that uses X11?

Apple has gone through various levels of support for it at various times.  I think the current behavior is to include stub libraries, and when any application actually tries to use it, a message pops up telling you to download and install XQuartz.

marscher:
I've added a freeglut recipe. Feel free to test it

jchodera:
Testing!

jchodera:
I get a freeglut build failure:
```
+ cmake -DCMAKE_INSTALL_PREFIX=/Users/choderaj/anaconda/envs/_build ..
CMake Warning at /Users/choderaj/anaconda/envs/_build/share/cmake-3.2/Modules/Platform/Darwin-Initialize.cmake:97 (message):
  CMAKE_OSX_DEPLOYMENT_TARGET is '10.6' but the matching SDK does not exist
  at:

   "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.6.sdk"

  Instead using SDK:

   "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk".
Call Stack (most recent call first):
  /Users/choderaj/anaconda/envs/_build/share/cmake-3.2/Modules/CMakeSystemSpecificInitialize.cmake:18 (include)
  CMakeLists.txt:2 (PROJECT)
```

jchodera:
I think you just need to specify some additional `CMAKE` variables for `osx` (just like OpenMM):
```
CMAKE_FLAGS+=" -DCMAKE_OSX_DEPLOYMENT_TARGET=10.9"
CMAKE_FLAGS+=" -DCMAKE_OSX_SYSROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk"
```

jchodera:
@marscher: How about you just add a `skip.yaml` to skip the `osx`, `win-32`, and `win-64` platforms for these packages, and then I can work on fixing the `osx` one in a separate PR?

marscher:
Done. Happy building on OSX!

jchodera:
Thanks!

rmcgibbo:
This doesn't build successfully on linux either: https://jenkins.choderalab.org/job/conda-omnia-release-linux-vagrant/175/consoleFull

PR #284, opened 2015-08-04
Fix windows packmol

### LINKED ISSUE #283: Should packmol.exe go to Anaconda\Scripts?
A clean installation of conda on windows adds the Anaconda and Anaconda\Scripts directories to the system path (but not Anaconda\bin). Packmol installs the packmol.exe executable to Anaconda\bin though. How about putting it into Anaconda\Scripts? This would eliminate the need of messing around with the path on Windows.

rmcgibbo:
Resolves #283

### DIFF
--- a/packmol/README.md
+++ b/packmol/README.md
@@ -1,3 +1,16 @@
 # packmol
 
-Note that `gfortran` must be installed to compile this.
+## Building Recipe
+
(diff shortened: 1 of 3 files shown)


rmcgibbo:
Also I built and pushed this from my machine.

PR #288, opened 2015-08-05
[pdbfixer] Update to pdbfixer 1.2

jchodera:


jchodera:
Not sure why it things appveyor failed---looks like it succeeded.

PR #291, opened 2015-08-09
[openmoltools] use openmm instead of openmm-dev

rmcgibbo:


mpharrigan:
You got rid of "openmm" instead of the "dev"

rmcgibbo:
Right, that's important.

PR #293, opened 2015-08-10
[bhmm] update to version 0.5.0

jchodera:


jchodera:
This works locally now too.

PR #305, opened 2015-08-23
[ambermini] fix static gfortran dependencies on osx

jchodera:


jchodera:
This works on my machine, but may need more testing.

rmcgibbo:
Making gcc a _runtime_ dependency on every platform seems like a heavy handed solution. Maybe only for osx?

jchodera:
Good idea.  The `gcc` package isn't even available for `win`.

rmcgibbo:
This works on my machine too.

jchodera:
Thanks!  Shouldn't affect the `linux` or `win` builds then.

PR #312, opened 2015-08-26
Revert "No jinja2/markupsafe on appveyor?"

rmcgibbo:
Latest release of conda-build (1.17.0) fixes the root cause of #242 upstream, so uninstalling jinja2 should not be necessary.

This reverts commit 2ed5a3e6ffd71644e90443714774b80a7e1c744a.


jchodera:
Awesome!

PR #341, opened 2015-10-09
WIP: cpptraj recipe

rmcgibbo:
@hainm, this successfully builds for me on linux (linking against system BLAS), and OS X using https://github.com/Amber-MD/cpptraj/pull/106 (linking against Accelerate for BLAS). Getting it to build on Windows using mingw-w64 will require more work.


hainm:
cool. +100.

hainm:
@rmcgibbo have you tested OSX on travis yet? I don't see your build.

rmcgibbo:
I've actually never used OS X travis.

rmcgibbo:
I just wrote and debugged the recipe locally on my OS X and linux machines. After this is merged, this this repository is tied to a dedicated build machine running Jenkins that @jchodera's group administers, so all the recipes get built and deployed to anaconda.org for OS X and linux through that infrastructure (and we use Appveyor for windows).

hainm:
I think travis only recently support osx. add ``os`` tag (work for parmed)

```bash
os:
    #- linux
    - osx
```

jchodera:
> I think travis only recently support osx. add os tag (work for ParmEd)

Wow, does that really work for everybody now?

swails:
> Wow, does that really work for everybody now?

@hainm just added a test env for ParmEd on OS X (see the apple symbol at the end of my most recent build: https://travis-ci.org/ParmEd/ParmEd

swails:
Ideally we'd like to get this deployed through the amber-md conda channel at some point.

Pinging @drroe in case he's curious/interested about this at all.

jchodera:
> @hainm just added a test env for ParmEd on OS X (see the apple symbol at the end of my most recent build: https://travis-ci.org/ParmEd/ParmEd

Awesome!  We should start adding these to our other projects.

I wonder if this also means we can start to eliminate Jenkins completely by migrating the conda package builds over to `travis` for `linux/osx`.

swails:
> I wonder if this also means we can start to eliminate Jenkins completely by migrating the conda package builds over to travis for linux/osx.

Would this work for the OpenCL and CUDA platforms in OpenMM?  My understanding was 'no' (unless Travis has an easy way of adding the CUDA toolkit version we want).

jchodera:
I think it is actually straightforward to *install* the toolkits (in the
root environment), just not test with them.

swails:
> I think it is actually straightforward to *install* the toolkits (in the
> root environment)

It's apparently been whitelisted: https://github.com/travis-ci/apt-package-whitelist/pull/596

But some caveats here:

- We would have no control over the CUDA version that gets installed if we go through apt
- It would add a ~1.5-2GB download *every* time conda packages were built (at minimum, if we didn't add it to the standard Travis test builds)

We *might* be able to alleviate the first point by downloading the toolkit from nVidia and installing it, assuming that was automatable somehow (keeping in mind that we have no root privileges in the container infrastructure).  And maybe we can cache the toolkit somewhere, but that might violate licensing terms in addition to any other issues with storing a very large file somewhere in/around Travis.

jchodera:
Is there a reason we couldn't use the same `rpm` installation mechanism for CUDA we use [in the current vagrant provisioning script](https://github.com/omnia-md/virtual-machines/blob/master/linux/setup_centos_vm.sh#L62-L69)?  Similarly, I have cached the AMD APP SDK privately, and use a mechanism to auto-install that.

swails:
Travis uses Ubuntu, so are `rpm`/`yum` even available?

If it could work, that'd be great.  Do you think it will be just as easy to get the OS X VM properly provisioned?

jchodera:
> Do you think it will be just as easy to get the OS X VM properly provisioned?

For Jenkins?  Or for travis?  I don't think there's anything that needs to be installed for travis, actually---OpenCL is a built-in part of the OS, isn't it?

swails:
> For Jenkins? Or for travis? I don't think there's anything that needs to be installed for travis, actually---OpenCL is a built-in part of the OS, isn't it?

Travis.

First step would be getting an OpenMM build working on OSX for Travis.  I'll see if I can port what @hainm did for ParmEd over there.

If we *can* get Travis building conda packages and uploading them, then we may be able to get rid of the Jenkins machines.

drroe:
@swails I am interested, thanks for the ping. I haven't made use of conda myself but it seems extremely useful.

swails:
I tried tweaking the Travis build script for OpenMM to run on the Mac.  Didn't have much luck, and I don't have time to try too hard, so I'm giving up.

hainm:
> I tried tweaking the Travis build script for OpenMM to run on the Mac. Didn't have much luck, and I don't have time to try too hard, so I'm giving up.

Me neither with cpptraj osx on travis.

PR #345, opened 2015-10-11
[WIP] Ambermini without OS X conda-gcc runtime dependency

rmcgibbo:
It took me a little while to figure this out, but I think it's correct now. Here's how to link a fortran binary executable or shared library on OS X without requiring that the user have any fortran-related binaries installed.

At the link step, link with GNU `gcc` and pass `-lm -static-libgcc /path/to/libgfortran.a /path/to/libquadmath.a`. Those paths can come from calling `$ gfortran -print-file-name=libgfortran.a`. `$ gfortran -print-file-name=libquadmath.a`

### DIFF
--- a/ambermini/build.sh
+++ b/ambermini/build.sh
@@ -1,25 +1,12 @@
 #!/bin/bash
-
-# Required for cbio cluster for some stupid reason.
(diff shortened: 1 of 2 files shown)


rmcgibbo:
This PR depends on https://github.com/choderalab/ambermini/pull/31 though.

PR #355, opened 2015-10-13
Update ParmEd conda version to 2.0.5 (stable)

swails:


rmcgibbo:
Reset the build number to 0

swails:
Why's that?

mpharrigan:
It's the first build for that version. Build number is for when the source doesn't change, but it has to be rebuilt for whatever reason

swails:
Got it.  Thanks!

PR #361, opened 2015-10-23
[bhmm] version bump 0.5.2

marscher:


marscher:
that was a quick merge!

Issue #370, opened 2015-11-02
Some Python3.3 binaries are missing


marscher:
For example the (new) bhmm release 0.5.2 has not been built for Python3.3, which blocks PyEMMA's Travis tests from passing, since the new version can not be installed for this Python branch.
```
(py33)marscher@jigsaw:~/workspace/pyemma$ conda search bhmm|grep 0.5.2
                             0.5.2                np19py35_0  omnia           
                             0.5.2                np19py34_0  omnia           
                             0.5.2                np19py27_0  omnia           
                             0.5.2               np110py35_0  omnia           
                          .  0.5.2               np110py34_0  omnia           
                          .  0.5.2               np110py27_0  omnia 
```

jchodera:
Hm.  It looks like the py33 package build isn't even being attempted:
https://jenkins.choderalab.org/job/conda-omnia-release-linux-vagrant/253/consoleFull

marscher:
But we used to have Python3.3, so why did we change this policy?

jchodera:
It appears that 33 has been eliminated from the default build list:
https://github.com/omnia-md/conda-recipes/blob/master/conda-build-all#L82

jchodera:
@rmcgibbo eliminated it in #351 when adding py35:
https://github.com/omnia-md/conda-recipes/commit/0bba21baa343fd6a82f344e34b54442f70abbf0a

I don't know if that was intended or not, but we should probably restore 33.

jchodera:
Can you create a PR to restore py33, @marscher?

rmcgibbo:
Nobody is using py33.

rmcgibbo:
Comtinuum isn't buidling new packages for py33 either. People that are using py3 have no barrier to upgrade to the latest minor releases.

jchodera:
By "nobody" do you mean "nobody in the field", or "none of our projects", because the latter is not correct.

In any case, support for major Python versions should probably be discussed before being dropped silently in a PR about something else.

rmcgibbo:
No *user* (or at the very least an insignificant number who have a roadblock-free upgrade path to py34 or py35) are using py33. The download counts are on anaconda.org. They're all ~ zero.

jchodera:
OK, so we should deprecate py33 across all projects.  We can totally do that---but as you can see, removing `py33` builds without warning has just broken lots of travis testing.

rmcgibbo:
```
$ conda search numpy 
[...]

PR #372, opened 2015-11-02
Specify versions of python supported

mpharrigan:


mpharrigan:
cc #370

jchodera:
+1 for documentation!

PR #374, opened 2015-11-05
Docker and OS X

rmcgibbo:
https://travis-ci.org/rmcgibbo/conda-recipes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,38 @@
+os:
+  - linux
+  - osx
+
+language: c
+
+sudo: required
+
+services:
+  - docker
+
+env:
+  global:
+    secure: "GjbAskNQDToMpTQ87p2TcG+JPdr+TIyWxarkzwd/KCQLuJqTKVVinWtKZwB+uaEnz2mmnXT7jFgbUU2Vuq0Dv78SOgTnqeh+ObP60imN4z8TB88PAFihv7JDJ45Nfbe7viKrWLWDYRebkObbJduc7xms0YEb9vs4e1+knTojiEc="
+
+install:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+      docker pull phusion/holy-build-box-64:latest;
+    fi
+
+script:
+  - if [[ "${TRAVIS_PULL_REQUEST}" == "true" && "${TRAVIS_BRANCH}" == "master" ]]; then
+      UPLOAD="--upload omnia";
+    else
+      UPLOAD=" ";
+    fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+      docker run -e UPLOAD -e BINSTAR_TOKEN
+          -t -i --rm -v `pwd`:/io phusion/holy-build-box-64:latest
+          bash /io/devtools/docker-build.sh;
+    elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      curl -s -O https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh;
+      bash Miniconda3-latest-MacOSX-x86_64.sh -b -p $HOME/anaconda;
+      export PATH=$HOME/anaconda/bin:$PATH;
+      conda config --add channels omnia;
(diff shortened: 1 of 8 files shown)


jchodera:
Whoa, what image is `phusion/holy-build-box-64`?

jchodera:
This is phenomenal!  Does this essentially replace our Jenkins setup for `linux` and `osx`?

rmcgibbo:
I learned about it from https://pmelsted.wordpress.com/2015/10/14/building-binaries-for-bioinformatics/
on twitter. See http://phusion.github.io/holy-build-box/

rmcgibbo:
I'm not really sure how it's different than just centos 5, exactly.

mpharrigan:
It has a newer gcc

rmcgibbo:
I don't think this has enough for OpenMM, because CUDA and OpenCL are a whole separate ball of wax, but otherwise I think it pretty much covers the same space as what the Jenkins servers do.

jchodera:
I've actually built a docker image with CUDA and the AMD APP SDK in it:
https://hub.docker.com/r/jchodera/docker-core-linux-build-environment/

The Dockerfile is [here](https://github.com/FoldingAtHome/docker-core-linux-build-environment).

jchodera:
For `linux`, could we just write a `Dockerfile` that starts with the `phusion/holy-build-box-64` and installs CUDA and the APP SDK, and have that automatically built by dockerhub?

That should just involve something like
```
FROM phusion/holy-build-box-64

# Install CUDA
ENV CUDA_HOME=/usr/local/cuda-7.0 CUDA_LIBPATH=/usr/local/cuda-7.0/lib64
ADD http://developer.download.nvidia.com/compute/cuda/repos/rhel6/x86_64/cuda-repo-rhel6-7.0-28.x86_64.rpm .
RUN rpm -i --quiet cuda-repo-rhel6-7.0-28.x86_64.rpm
RUN yum clean expire-cache
RUN yum install -y --quiet cuda

# Install AMD SDK for OpenCL
ENV OPENCL_HOME=/opt/AMDAPPSDK-2.9-1 OPENCL_LIBPATH=/opt/AMDAPPSDK-2.9-1/lib/x86_64
ADD http://jenkins.choderalab.org/userContent/AMD-APP-SDK-linux-v2.9-1.599.381-GA-x64.tar.bz2 .
RUN tar xjf AMD-APP-SDK-linux-v2.9-1.599.381-GA-x64.tar.bz2
RUN ./AMD-APP-SDK-v2.9-1.599.381-GA-linux64.sh -- -s -a yes
```

rmcgibbo:
Can you try doing that? I've never used dockerhub. Alternatively, could we just use that image instead? I don't remember if we want build off centos 5 or centos 6.

jchodera:
Happy to try.  Give me a moment.

I believe we wanted to build off of CentOS 5 if at all possible, so I"m excited about also using the `phusion/holy-build-box-64` image for OpenMM F@h cores as well.

rmcgibbo:
Yeah, that sounds good. I might merge this first, since this issue only
applies for OpenMM.

jchodera:
Please do!

Is there a way to get it to skip OpenMM?

jchodera:
I don't think it's possible to easily install CUDA on centos 5.x, but I realize that our `Vagrant` build VM was based on [centos 6.6](https://github.com/omnia-md/virtual-machines).

We might therefore be able to get by with using a tiny modification of the [fah openmm-core build docker image](https://hub.docker.com/r/jchodera/docker-core-linux-build-environment/) which is also based on centos 6.6.

I'll explore switching the docker image in a separate PR.

rmcgibbo:
We could just use a separate CentOS 6 / CUDA image for OpenMM and the Centos 5 image for everything else. OpenMM is a special flower...

jchodera:
That could work too, but would it be a separate travis branch, or a separate github repo?

rmcgibbo:
Separate travis branch, probably.

rmcgibbo:
Can you pause the Jenkins server? I want to file a new PR that bumps the build number on a bunch of recipes to make sure the Travis stuff is working, and don't want Jenkins and Travis to race each other.

jchodera:
Paused!

PR #375, opened 2015-11-05
Updates

rmcgibbo:
Misc updates to a lot of recipes, and bumping of the build number. Mostly I want to trigger the new travis/docker build to make sure everything runs. Also I noticed that some packages were installing as eggs, which is kind of janky. Using `pip install .` instead of `python setup.py install` fixes that.

### DIFF
--- a/alchemy/build.sh
+++ b/alchemy/build.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-$PYTHON setup.py install
+pip install .
--- a/alchemy/meta.yaml
+++ b/alchemy/meta.yaml
@@ -8,8 +8,7 @@ source:
 
 
 build:
-  preserve_egg_dir: True
-  number: 0
+  number: 1
   skip:
      - [py3k]
 
--- a/behave/meta.yaml
+++ b/behave/meta.yaml
@@ -6,25 +6,10 @@ source:
   fn: behave-1.2.5.tar.gz
   url: https://pypi.python.org/packages/source/b/behave/behave-1.2.5.tar.gz
   md5: 3af4bc2886d2be982e41bfaba61da3bc
-#  patches:
-   # List any patch files here
-   # - fix.patch
 
 build:
-  #preserve_egg_dir: True
   entry_points:
-    # Put any entry points (scripts to be generated automatically) here. The
-    # syntax is module:function.  For example
-    #
-    # - behave = behave:main
-    #
-    # Would create an entry point called behave that calls behave.main()
-
     - behave = behave.__main__:main
-
-  # If this is a new build for the same version, increment the build
-  # number. If you do not include this key, it defaults to 0.
-  # number: 1
 
 requirements:
   build:
--- a/devtools/docker-build.sh
+++ b/devtools/docker-build.sh
@@ -4,6 +4,7 @@ set -e
 source /hbb_exe/activate
 
 set -x
+yum install -y git
 curl -s -O https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh
 bash Miniconda3-latest-Linux-x86_64.sh -b -p /anaconda
 PATH=/opt/rh/devtoolset-2/root/usr/bin:/opt/rh/autotools-latest/root/usr/bin:/anaconda/bin:$PATH
--- a/docopt/build.sh
+++ b/docopt/build.sh
@@ -1,9 +1,3 @@
 #!/bin/bash
 
-$PYTHON setup.py install
-
-# Add more build steps here, if they are necessary.
-
-# See
-# http://docs.continuum.io/conda/build.html
-# for a list of environment variables that are set during the build process.
+pip install .
--- a/docopt/meta.yaml
+++ b/docopt/meta.yaml
@@ -6,24 +6,9 @@ source:
   fn: docopt-0.6.2.tar.gz
   url: https://pypi.python.org/packages/source/d/docopt/docopt-0.6.2.tar.gz
   md5: 4bc74561b37fad5d3e7d037f82a4c3b1
-#  patches:
-   # List any patch files here
-   # - fix.patch
 
-# build:
-  #preserve_egg_dir: True
-  #entry_points:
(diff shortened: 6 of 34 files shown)


jchodera:
Why is installing as an egg janky?

rmcgibbo:
I got the idea mostly from https://mail.scipy.org/pipermail/numpy-discussion/2015-October/074086.html

jchodera:
Everything build OK?

rmcgibbo:
Here are some of the commands that seemed to work to get CUDA and TeX installed on the CentOS 5 image, that could probably be made into a new image.
```
# three necessary files
curl -s -O http://ctan.mackichan.com/systems/texlive/tlnet/install-tl-unx.tar.gz
curl -s -O http://download.fedoraproject.org/pub/epel/5/x86_64/epel-release-5-4.noarch.rpm
curl -s -O http://developer.download.nvidia.com/compute/cuda/7_0/Prod/local_installers/rpmdeb/cuda-repo-rhel6-7-0-local-7.0-28.x86_64.rpm

rpm -ivh epel-release-5-4.noarch.rpm
yum install -y --quiet dkms libvdpau git wget

# Install TeX
tar -xzf install-tl-unx.tar.gz
cd install-tl-*
./install-tl -profile /io/devtools/texlive.profile
cd -
rm -rf install-tl-unx.tar.gz install-tl-*
export PATH=/usr/local/texlive/2015/bin/x86_64-linux:$PATH

rpm -i cuda-repo-rhel6-7-0-local-7.0-28.x86_64.rpm
yum clean expire-cache
yum install -y --quiet cuda-core-7-0-7.0-28.x86_64 cuda-cufft-dev-7-0-7.0-28.x86_64 cuda-cudart-dev-7-0-7.0-28.x86_64
rpm --nodeps -Uvh /var/cuda-repo-7-0-local/xorg-x11-drv-nvidia-libs-346.46-1.el6.x86_64.rpm
rpm --nodeps -Uvh /var/cuda-repo-7-0-local/xorg-x11-drv-nvidia-devel-346.46-1.el6.x86_64.rpm
ln -s /usr/include/nvidia/GL/  /usr/local/cuda-7.0/include/
yum clean -y all
```

rmcgibbo:
Here's the texlive file: https://gist.github.com/rmcgibbo/20038984e79c0176d539

mpharrigan:
Why do you need tex for cuda?

jchodera:
TeX is necessary for building `openmm`.

Thanks for the CUDA tips!  I'll try some more experimentation there.

rmcgibbo:
I tried getting your image (`jchodera/centos5-build-box`) from dockerhub.
 - We don't need the extra `boost`. Packages that depend on boost can get it through conda.
 - All of the tarballs should not live in the final image (`/AMD-APP-SDK-v2.9-1.599.381-GA-linux64.sh`, `/AMD-APP-SDK-linux-v2.9-1.599.381-GA-x64.tar.bz2`, `/epel-release-5-4.noarch.rpm`, `/cuda-repo-rhel6-7-0-local-7.0-28.x86_64.rpm`, `/boost_1_55_0.tar.bz2`). Can we delete them after installing
  * I think `rm -rf /var/cuda-repo-7-0-local/ /var/cache/yum/cuda-7-0-local/`
  * And get rid of `/opt/AMDAPPSDK-2.9-1/samples/`

 - I don't think `TeX` got installed correctly; `/usr/local/texlive/` does not exist.
    * You probably need to `ADD` the texlive profile file, or at least mount it.

jchodera:
Thanks!  I had tried to clean up the tarballs but didn't get all of them removed.

rmcgibbo:
Everything's working now with [this image](https://hub.docker.com/r/rmcgibbo/centos5-build-box/). I believe there are no remaining consequential differences from the Jenkins setup w.r.t. linux. For OS X, the travis environment doesn't have CUDA or TeX, so the OpenMM binaries aren't built with the CUDA platform or the pdf manual. AFAICT, there's no headless method to install CUDA on the Travis OS X VMs.

jchodera:
Oh, I hadn't realized you fixed it!

rmcgibbo:
I should probably move the Dockerfile into this repo.

jchodera:
> For OS X, the travis environment doesn't have CUDA or TeX, so the OpenMM binaries aren't built with the CUDA platform or the pdf manual. AFAICT, there's no headless method to install CUDA on the Travis OS X VMs.

Wait, we don't support CUDA on `osx`, do we?  We just support `Reference`, `CPU`, and `OpenCL`.

rmcgibbo:
Oh, that's convenient then. I didn't know that.

jchodera:
No, I'm wrong---the build environment for `osx` does generally include CUDA:
https://github.com/pandegroup/openmm/wiki/OpenMM-build-platforms-and-environments

Maybe we can install CUDA on a local machine and just `tar` up the files it unpacks?

jchodera:
> I should probably move the Dockerfile into this repo.

That would be great!  We can also use this as the basis for building more-broadly-compatible openmm cores for F@h.

jchodera:
> We don't need the extra boost. Packages that depend on boost can get it through conda.

The `boost` was for the F@h `openmm-core`, BTW.

jchodera:
Or we could potentially use [this scheme](http://apple.stackexchange.com/questions/73926/is-there-a-command-to-install-a-dmg) to install the CUDA `dmg` through the command-line:
```
sudo hdiutil attach <image>.dmg
sudo installer -package /Volumes/<image>/<image>.pkg -target /
sudo hdiutil detach /Volumes/<image>
```
Or do we not have `sudo` in the `osx` travis enviroment?

rmcgibbo:
I tried that, but it doesn't work. It's not a .pkg, it's a gui installer, and you can't even use the enter key, you need to move the mouse to the button.

rmcgibbo:
Actually, the network installer downloads the whole toolkit from this link: `http://developer.download.nvidia.com/compute/cuda/7_0/Prod/network_installers/mac/cuda_mac_installer_tk.tar.gz`, so it might be possible...

Issue #381, opened 2015-11-07
Add RDKit


rmcgibbo:
Assigning @rbharath :)

jchodera:
Isn't there already [an official recipe](https://github.com/rdkit/conda-rdkit/tree/master/rdkit) and [an official conda package](https://anaconda.org/rdkit/rdkit)?

rmcgibbo:
Yeah, so it should be easy. But it hard-depends on a different boost than the one here, etc.

jchodera:
To be clear: We need to build our own recipe because the [official RDKit conda package](https://anaconda.org/rdkit/rdkit) depends on different library versions than the ones we use on omnia?

rmcgibbo:
Yes, I think that's correct. Mostly I was kind of ribbing @rbharath with this issue, because he was saying that this was on his agenda. I don't know all the technical details.

jchodera:
Was asking just because I was talking with some folks yesterday who wondered what happens when we run into an issue where two different sets of conda-installed tools are linked to different versions of the same dynamic library.

rmcgibbo:
- If the packages fully declare their dependencies, like "X depends on lib==1.5", "Y depends on lib==1.6", and you try to install both X and Y into the same environment, the resolver will say there's a conflict (it just solves a SAT problem), and refuse.
 - If one packages fully declares its dependencies but the other doesn't, like "X depends on lib", "Y depends on lib==1.6", then it'll install both. But if X _actually_ has a binary dependency on lib==1.5, and the library isn't binary forward compatible (very rare), then you'll get into a broken state.

jchodera:
Thanks!  I guess the proper way to do this is to explicitly declare those dependencies as "X depends on lib>=1.5" and make sure any dynamic library dependencies are appropriately packaged as conda packages, rather than distributed along with the conda package for the tool linking them.

rmcgibbo:
For boost, the real solution is just to get a proper boost package in the default channels.

rbharath:
@jchodera: There are also a couple other issues with the standard conda-rdkit build besides boost library version. For example, there's a test that I have to manually disable each time I build to get it to work out right.

marscher:
currently rdkit blocks travis osx builds because it needs more than 40 min (maximum wall time of travis). Maybe try invoking make with at least two jobs as it might speed things up.

PR #384, opened 2015-11-07
RDKit Recipe

rbharath:
This PR adds a recipe for rdkit makes some minor modifications to the standard conda-rdkit recipe at https://github.com/rdkit/conda-rdkit (addresses #381). Namely,

1. The hard-coded requirement for boost == 1.56.0 is removed.
2. A passing test suite is no longer required for a successful build (I've noticed that one test out of 81 can be quite flaky, and I've been disabling it in practice)

### DIFF
--- a/rdkit/build.sh
+++ b/rdkit/build.sh
@@ -23,6 +23,6 @@ cmake \
     -D CMAKE_BUILD_TYPE=Release \
     .
 
-make
+make -j $CPU_COUNT
 
 make install
--- a/rdkit/build.sh
+++ b/rdkit/build.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+$PYTHON "$RECIPE_DIR/fetch_inchi.py"
+$PYTHON "$RECIPE_DIR/fetch_avalontools.py"
+PY_INC=`$PYTHON -c "from distutils import sysconfig; print (sysconfig.get_python_inc(0, '$PREFIX'))"`
+
+cmake \
+    -D RDK_INSTALL_INTREE=OFF \
+    -D RDK_INSTALL_STATIC_LIBS=OFF \
+    -D RDK_BUILD_INCHI_SUPPORT=ON \
+    -D RDK_BUILD_AVALON_SUPPORT=ON \
+    -D RDK_USE_FLEXBISON=OFF \
+    -D RDK_BUILD_THREADSAFE_SSS=ON \
+    -D RDK_TEST_MULTITHREADED=ON \
+    -D AVALONTOOLS_DIR=$SRC_DIR/External/AvalonTools/src/SourceDistribution \
+    -D CMAKE_SYSTEM_PREFIX_PATH=$PREFIX \
+    -D CMAKE_INSTALL_PREFIX=$PREFIX \
+    -D Python_ADDITIONAL_VERSIONS=${PY_VER} \
+    -D PYTHON_EXECUTABLE=$PYTHON \
+    -D PYTHON_INCLUDE_DIR=${PY_INC} \
+    -D PYTHON_NUMPY_INCLUDE_PATH=$SP_DIR/numpy/core/include \
+    -D BOOST_ROOT=$PREFIX -D Boost_NO_SYSTEM_PATHS=ON \
+    -D CMAKE_BUILD_TYPE=Release \
+    .
+
+make
+
+make install
--- a/rdkit/meta.yaml
+++ b/rdkit/meta.yaml
@@ -11,6 +11,8 @@ source:
 
 build:
   number: 0 
+  skip:
+    - [osx or win]
 
 requirements:
   build:
--- a/rdkit/meta.yaml
+++ b/rdkit/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: rdkit
-  version: 2015.03.1
+  version: 2015.09.1
 
 source:
   git_url: https://github.com/rdkit/rdkit.git
-  git_tag: Release_2015_03_1 
+  git_tag: Release_2015_09_1
   patches:
     - rdpaths.patch
     - rdconfig.patch [win]
--- a/rdkit/meta.yaml
+++ b/rdkit/meta.yaml
@@ -27,9 +27,9 @@ requirements:
     - python
     - numpy
 
-#test:
-#  imports:
-#    - rdkit
+test:
+  imports:
+    - rdkit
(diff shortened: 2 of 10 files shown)


rbharath:
The Travis-CI builds fail because they run out of time. @rmcgibbo @jchodera: Have either of you run into similar issues in the past?

rdkit/build.sh
```diff
@@ -0,0 +1,28 @@
...
+    -D CMAKE_INSTALL_PREFIX=$PREFIX \
+    -D Python_ADDITIONAL_VERSIONS=${PY_VER} \
+    -D PYTHON_EXECUTABLE=$PYTHON \
+    -D PYTHON_INCLUDE_DIR=${PY_INC} \
+    -D PYTHON_NUMPY_INCLUDE_PATH=$SP_DIR/numpy/core/include \
+    -D BOOST_ROOT=$PREFIX -D Boost_NO_SYSTEM_PATHS=ON \
+    -D CMAKE_BUILD_TYPE=Release \
+    .
+
+make
```
rmcgibbo: try adding -j$CPU_COUNT

rdkit/meta.yaml
```diff
@@ -0,0 +1,36 @@
+package:
+  name: rdkit
+  version: 2015.03.1
+
+source:
+  git_url: https://github.com/rdkit/rdkit.git
```
rmcgibbo: can you change this to pull from the https:// for the release tarball from https://github.com/rdkit/rdkit/releases instead of git?

rmcgibbo:
I haven't had a build timeout like this before.
 - It looks like it _almost_ finished -- it was tarring up the final result when it died, I think, so maybe it'll just get lucky and finish next time.
 - Adding the parallel make flag might also be enough to get it over the edge.
 - Alternatively, you might be able to package the AvalonTools and Inchi separately, and then have then as build-dependencies, so that each build was lighter. But that's a hassle.
 - The build could be run manually and uploaded. With the new linux docker setup, you get the same exact build environment if you run [this command](https://github.com/omnia-md/conda-recipes/blob/master/.travis.yml#L29-L31) on your machine, or if it runs on travis, since it's containerized.

rdkit/meta.yaml
rbharath: git_tag is specified below, so it is pulling from the release version and not head.

rbharath:
Ok, I've changed make to use ```make -j $CPU_COUNT```. I'm hoping this caries the build through travis. If it doesn't, then I'll move this PR over to docker.

rdkit/meta.yaml
rmcgibbo: i know. but if you look at the windows log, git is failing. and git is slow, because it has to clone the whole history, as opposed to just getting the tarball over http

rmcgibbo:
Here's the OS X linker failure.

```
Undefined symbols for architecture x86_64:

  "boost::archive::basic_text_iprimitive<std::istream>::~basic_text_iprimitive()", referenced from:

      RDKit::FilterCatalogParams::initFromStream(std::istream&) in FilterCatalog.cpp.o

      RDKit::FilterCatalog::FilterCatalog(std::string const&) in FilterCatalog.cpp.o

      boost::archive::text_iarchive::~text_iarchive() in FilterCatalog.cpp.o

      boost::archive::text_iarchive::~text_iarchive() in FilterCatalog.cpp.o

      RDKit::FilterCatalogEntry::initFromStream(std::istream&) in FilterCatalogEntry.cpp.o

      boost::archive::text_iarchive::~text_iarchive() in FilterCatalogEntry.cpp.o

      boost::archive::text_iarchive::~text_iarchive() in FilterCatalogEntry.cpp.o

      ...

  "boost::archive::basic_text_oprimitive<std::ostream>::~basic_text_oprimitive()", referenced from:

      RDKit::FilterCatalogParams::toStream(std::ostream&) const in FilterCatalog.cpp.o

      RDKit::FilterCatalog::Serialize() const in FilterCatalog.cpp.o

      boost::archive::text_oarchive::~text_oarchive() in FilterCatalog.cpp.o

      boost::archive::text_oarchive::~text_oarchive() in FilterCatalog.cpp.o

      RDKit::FilterCatalogEntry::toStream(std::ostream&) const in FilterCatalogEntry.cpp.o

      boost::archive::text_oarchive::~text_oarchive() in FilterCatalogEntry.cpp.o

      boost::archive::text_oarchive::~text_oarchive() in FilterCatalogEntry.cpp.o

      ...

ld: symbol(s) not found for architecture x86_64

clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

rbharath:
@rmcgibbo: Where did you get the failing linker output?

The travis link says that the OS X build ran, while the linux build again runs out of time (see https://travis-ci.org/omnia-md/conda-recipes/builds/90001408).  The linux build seems to fail since the rdkit test tries to rebuild rdkit(!). I'll try disabling tests and seeing if the build completes

rmcgibbo:
Click "download log" near the upper right of the output box. Links to: https://s3.amazonaws.com/archive.travis-ci.org/jobs/90001410/log.txt

rmcgibbo:
No, the test did not try to rebuild RDKit. Actually the build looks like it worked just fine on linux.

rmcgibbo:
The apparent problem is just that the PR is trying to build the recipe for multiple different versions of Python (27, 34, 35). It finished the first build, but couldn't make it through the second or third.

rmcgibbo:
And that's totally okay, because we can merge it, and I can just hit "restart" a couple times, and within a day, we'll have all of the versions built and uploaded.

rbharath:
Ah, I see. In that case, let me re-enable tests for rdkit before we merge changes. 

For the MAC linker error, this seems to be linked to boost 1.59.0 (see https://trac.macports.org/ticket/48717). I don't have a MAC handy, so this one is a little tough to debug though.

rmcgibbo:
Okay. Can you just disable mac and windows then? The syntax is like this: https://github.com/omnia-md/conda-recipes/blob/master/subprocess32/meta.yaml#L10-L12

rbharath:
Ok, I've disabled osx/windows builds for now. It would be nice to get these back in later though, but I'll have to get access to an OSX/Windows machine for testing first.

rmcgibbo:
Yeah. The windows one would be (in part, at least) fixed by pulling from http, not git, but there could be other downstream problems too.

PR #385, opened 2015-11-08
use mdtraj<1.5 to avoid bugs due to removed private api

marscher:


franknoe:
Why do we need this? PR #614 in PyEMMA actually requires mdtraj1.5. If we release 2.0.2 that's it.

Or do you expect further delays for a 2.0.2 release? In that case, we still have a problem with the pip install which is not fixed by this recipe update

franknoe:
So I think we need to release 2.0.2 asap in order to keep the number of broken installations small.

franknoe:
ok, merging this to have at least a preliminary solution

PR #391, opened 2015-11-12
[pymbar] pymbar 3.0.0.beta2

jchodera:


rmcgibbo:
```
Processing /Users/travis/anaconda/conda-bld/work/pymbar-3.0.0.beta2
    Complete output from command python setup.py egg_info:
      File "<string>", line 20, in <module>
      File "/tmp/pip-nDsZB5-build/setup.py", line 14, in <module>
        import six
    ImportError: No module named six
```

jchodera:
Thanks!  I just saw that too.

I wonder if there is a way to have the travis script signal *failure* in a PR if a package build fails, but to ignore failures in actual commits (to make sure all packages get built).

rmcgibbo:
It's probably possible.

rmcgibbo:
Also looks like the tests aren't running.

jchodera:
Hm. Do we really want tests to run if they take ~10 min?

jchodera:
Disabling the tests since they take too long here.  Each package should be verifying the installed package works correctly with `devtools` anyway.

PR #395, opened 2015-11-14
[autograd] Add autograd recipe via `conda skeleton pypi autograd`

jchodera:


rmcgibbo:
It looks like it's building a C extension against numpy, so I think it needs `numpy x.x`

jchodera:
I need `numpy x.x` in *addition* to the `numpy >=1.9` dependency?

rmcgibbo:
Just remove the `>= 1.9`. We only build with 1.9 and 1.10 anyways.

PR #396, opened 2015-11-17
[progressbar] Add progressbar from conda/conda-recipes

jchodera:
From https://github.com/conda/conda-recipes/tree/master/progressbar, since somehow that package is no longer be maintained in the main anaconda channel.

### DIFF
--- a/progressbar/bld.bat
+++ b/progressbar/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
--- a/progressbar/build.sh
+++ b/progressbar/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install
+
+# Add more build steps here, if they are necessary.
+
+# See
+# http://docs.continuum.io/conda/build.html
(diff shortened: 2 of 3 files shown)


jchodera:
Ah. This may be because `progressbar` is not yet py3.x ready:
```
+ /Users/travis/anaconda/envs/_build/bin/python setup.py install
  File "setup.py", line 5, in <module>
    import progressbar
  File "/Users/travis/anaconda/conda-bld/work/progressbar-2.3/progressbar/__init__.py", line 59, in <module>
    from progressbar.widgets import *
  File "/Users/travis/anaconda/conda-bld/work/progressbar-2.3/progressbar/widgets.py", line 121, in <module>
    class FileTransferSpeed(Widget):
  File "/Users/travis/anaconda/envs/_build/lib/python3.4/abc.py", line 133, in __new__
    cls = super().__new__(mcls, name, bases, namespace)
ValueError: 'format' in __slots__ conflicts with class variable
Command failed: /bin/bash -x -e /Users/travis/build/omnia-md/conda-recipes/progressbar/build.sh
Removing old build environment
Removing old work directory
```

rmcgibbo:
Can you add a `skip` entry then?

jchodera:
Can we just unmerge?

swails:
> Can we just unmerge?

:).  Why not just revert?  That's easy to do from `gitk`

bas-rustenburg:
@jchodera: Is progressbar2 py3.x ready? You could consider switching to that. http://progressbar-2.readthedocs.org/en/latest/

PR #397, opened 2015-11-17
Revert "[progressbar] Add progressbar from conda/conda-recipes"

rmcgibbo:
Reverts omnia-md/conda-recipes#396


rmcgibbo:
I assume there's probably another well-maintained progressbar popular library, since it's used in many apps.

jchodera:
Thanks! And agree.

PR #416, opened 2016-01-09
Skip building sphinxcontrib packages on win

jchodera:
Apparently, we [can't build `noarch` packages on `win`](https://ci.appveyor.com/project/rmcgibbo/conda-recipes/build/1.0.572/job/fp2wviopka3jae69#L390):
```
[noarch_python] Error: Python noarch packages can currently not be created on Windows systems.
```
This PR skips builds of the new `sphinxcontrib-*` packages on `win`.


jchodera:
Though I wonder if the `noarch` thing is the reason these packages are not being uploaded.  They seem to be the only ones with `noarch_python: True`.

PR #417, opened 2016-01-09
Fix uploading of noarch-python packages

### LINKED ISSUE #415: conda-build-all not uploading built packages
Not sure what's going on here, but the new `sphinxcontrib-*` packages are not being uploaded after being built:
https://travis-ci.org/omnia-md/conda-recipes/jobs/101220623

rmcgibbo:
Resolves #415

### DIFF
--- a/conda-build-all
+++ b/conda-build-all
@@ -6,6 +6,7 @@ import operator
 import glob
 import os
 import time
+import tarfile
 import os.path
 import argparse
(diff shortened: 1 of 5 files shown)


rmcgibbo:
Note that there is a race condition when we have multiple build workers all trying to build noarch python packages and upload them. One of them will be first, but I'm not sure what will happen to the others.

jchodera:
Yay!  Thanks!

PR #419, opened 2016-01-13
Fix issue with forcing re-compile of packages that already exist on omnia channel

jchodera:


jchodera:
Let's make sure this works before merging...

PR #424, opened 2016-01-28
Build pyhmc on windows

mpharrigan:
Let's see if this works / why it doesn't work

### DIFF
--- a/pyhmc/meta.yaml
+++ b/pyhmc/meta.yaml
@@ -9,6 +9,8 @@ source:
 
 build:
   number: 1
+  skip:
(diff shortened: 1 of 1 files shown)


mpharrigan:
Looks like it can't do numpy 1.9

How do I use that in the "skip" field

mpharrigan:
Going based off of https://github.com/conda/conda-build/blob/master/conda_build/metadata.py#L38

Couldn't find it in the documentation

mpharrigan:
Yay! should be good to merge

mpharrigan:
bump

PR #427, opened 2016-01-29
pyemma version bump 2.0.3

marscher:


marscher:
test win32-py27 failed because of #428

PR #429, opened 2016-01-30
Adds deepchem, keras, and theano

rbharath:
Adds conda-recipe for deepchem (http://deepchem.io). deepchem requires keras and theano to build, so adds those recipes as well.


jchodera:
Cool!

rbharath:
Ping @rmcgibbo. Could you take a quick look at this PR? There's a strange failure due to GLIBC

    ImportError: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.9' not found (required by /root/.python-eggs/openbabel-1.8.3-py2.7-linux-x86_64.egg-tmp/_openbabel.so)

conda-build-all
```diff
@@ -256,8 +258,9 @@ def required_builds(metadatas, pythons, numpys, channel_urls, verbose):
     if not os.path.isdir(conda_build_config.bldpkgs_dir):
         os.makedirs(conda_build_config.bldpkgs_dir)
     conda_build_index.update_index(conda_build_config.bldpkgs_dir)
-    index = conda_build_build.get_build_index(
-        clear_cache=True, channel_urls=channel_urls, override_channels=True)
+    index = conda.api.get_index(
```
mpharrigan: Why did you change this?

rbharath: This is a general fix (not just for deepchem) conda_build was updated a couple days ago and changed the get_build_index function to not accept channel_urls and override_channels function. Robert and I pieced together a reasonable looking fix.

rmcgibbo: this commit: https://github.com/conda/conda-build/commit/c61983153dff55819b92ca793f40f62921a4dca1 removed the kwargs we were using

rmcgibbo:
try bumping the `number` field up by one in the openbabel recipe here https://github.com/omnia-md/conda-recipes/blob/master/openbabel/meta.yaml#L15 in this PR to force that recipe to be rebuilt (it might have been built incorrectly, and see if that fixes it)

rmcgibbo:
It looks like there's still some kind of failure running your test suite, but everything imports correctly, including, I think, openbabel.

theano/meta.yaml
```diff
@@ -0,0 +1,68 @@
+package:
+  name: theano
+  version: "0.7.2"
+
+source:
+  git_url: https://github.com/Theano/Theano/
+
```
rmcgibbo: you should probably put a specific git hash here, for the current master branch.

rmcgibbo:
but I would be fine to merge this now, especially with the fix to `conda-build-all` and the apparent openbabel fix.

mpharrigan:
Won't it not push the package if it thinks the build (/tests) failed?

rmcgibbo:
probably not, but it'll still push the openbabel/keras/theano, so the remaining fixes will be quick.

mpharrigan:
ah! thought the problems was in one of the dependencies for some reason..

rmcgibbo:
@rbharath: merging this, but as matt pointed out the deepchem part will still probably not get pushed to anaconda.org due to failure there.

rbharath:
@rmcgibbo: The tests that fail write some data to a tempfile and then read it later. The reads from the tempfile fail. This is weird since these tests pass consistently on my local machine. Do you have any thoughts on what could account for the difference? What's a good way to debug these sorts of build failures?

rmcgibbo:
I'm not sure. You can reproduce the exact environment though that's used in the cloud by installing Docker and then running the code from .travis.yml through docker.

rmcgibbo:
i.e. try running
```
docker run -i -t --rm -v `pwd`:/io rmcgibbo/centos5-build-box:latest /bin/bash
```

jchodera:
Looks like whatever you guys did to `conda-build-all` broke [`conda-dev-recipes`](https://github.com/omnia-md/conda-dev-recipes).  Is there a way we can make sure that doesn't happen?

jchodera:
https://travis-ci.org/omnia-md/conda-dev-recipes/jobs/105829088#L227-L237
```
+ /io/conda-build-all --force --upload omnia /io/README.md /io/appveyor.yml /io/conda-build-all /io/devtools /io/openmm-dev /io/parmed-dev
  File "/io/conda-build-all", line 353, in <module>
    sys.exit(main())
  File "/io/conda-build-all", line 97, in main
    return execute(args, p)
  File "/io/conda-build-all", line 323, in execute
    to_build = list(required_builds(metadatas, pythons, numpys, channel_urls, verbose=args.verbose))
  File "/io/conda-build-all", line 260, in required_builds
    clear_cache=True, channel_urls=channel_urls, override_channels=True)
TypeError: get_build_index() got an unexpected keyword argument 'channel_urls'
```

jchodera:
Any chance you guys can fix?

jchodera:
Will try rerunning travis just in case you already did...

jchodera:
Looks like travis passes now!

PR #454, opened 2016-02-27
gpy

cxhernandez:
adding recipe for [GPy](https://github.com/SheffieldML/GPy)

### DIFF
--- a/gpy/bld.bat
+++ b/gpy/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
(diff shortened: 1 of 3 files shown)


cxhernandez:
Is there a way to list pip dependencies in meta.yaml or would I have to make a conda build for `paramz` as well?

jchodera:
No, sadly there isn't.  Check out "conda skeleton"---it lets you create a conda recipe from a pypi package.  You can then add the recipe here

jchodera:
Just retriggered travis now that #455 is merged.

PR #456, opened 2016-03-03
[alchemy] 1.1.4

jchodera:


jchodera:
AppVeyor error seems to be with `gpy`, so created #457

PR #459, opened 2016-03-05
[ambermini and packmol] Update ambermini and packmol builds to fix `GFORTRAN_1.4' not found 

jchodera:
cc: https://github.com/choderalab/openmoltools/pull/198#issuecomment-192029702

### DIFF
--- a/ambermini/meta.yaml
+++ b/ambermini/meta.yaml
@@ -17,9 +17,11 @@ requirements:
     - bison    [unix]
     - gcc      [unix]
     - mingwpy  [win]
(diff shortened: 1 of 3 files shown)


jchodera:
This is a hunch, but I think rebuilding these packages with current `gcc` and `libgcc` dependencies may fix this.  If not, we will add `libgfortran` dependencies as suggested in https://github.com/choderalab/openmoltools/pull/198#issuecomment-192029702

jchodera:
Actually, it probably doesn't hurt to add those dependencies anyway.

PR #463, opened 2016-03-06
add osprey

cxhernandez:
built with conda-build skeleton


cxhernandez:
appveyor seems to build `osprey` just fine (still failing on `gpy`)

jchodera:
Great!

PR #458, opened 2016-03-03
try peastman suggestion

cxhernandez:
re: https://github.com/omnia-md/conda-recipes/issues/457

PR #465, opened 2016-03-09
Fix gpy stuff

rmcgibbo:
This should fix the gpy problems from #458

### DIFF
--- a/gpy/0001-Add-basesetup.py.patch
+++ b/gpy/0001-Add-basesetup.py.patch
@@ -1,0 +1,366 @@
+From 9f246794a15d6e09960a03c5f47be545c774d551 Mon Sep 17 00:00:00 2001
+From: "Robert T. McGibbon" <rmcgibbo@gmail.com>
+Date: Wed, 9 Mar 2016 10:24:33 -0800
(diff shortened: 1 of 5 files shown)


rmcgibbo:
I changed back to pulling from the `devel` git branch of GPy, because python3 was broken on the latest release. But I updated the version number to 0.8.8.post1, since it's beyond 0.8.8. This PR also includes a fixed version of the Windows OpenMP compiler detection code.

jchodera:
Wow, that's a lot of new code!

rmcgibbo:
It looks like that, but it's not. That code is just copied from mdtraj for detecting whether we're using Visual Studio and passing the right flags.

rmcgibbo:
(https://github.com/mdtraj/mdtraj/blob/master/basesetup.py)

jchodera:
Looks like there is [a linux failure on travis](https://travis-ci.org/omnia-md/conda-recipes/jobs/114870980#L671-L677), but awesome work otherwise! Thanks!

rmcgibbo:
Ready to merge

jchodera:
Awesome, thanks!

Issue #467, opened 2016-03-17
Shiftx2 got sent to "pre" label


mpharrigan:
I haven't been keeping up with this, but I would hope that new recipes would be put in the main label by default.

cc #466 

Why did this get sent to the "pre" label?

mpharrigan:
cc #445

mpharrigan:
I'm manually changing the label from pre to main on each of the bz2 files

jchodera:
Tagging @marscher : Any idea what happened here?  Did you make some changes to the build script that may have caused this?

marscher:
I think so. The check for a package to go to "pre" is a setuptools related version class attribute name "pre-release", which in this case is true for a version number tagged "A" for alpha. So this is actually expected behaviour for this kind over version number scheme. How do we want to tag releases to be public or restrict them to testing community? 
This question arises because we somehow need to automatize how to treat pre-releases.

mpharrigan:
Ok, shiftx2 doesn't follow pep440 at all, so we'll just have to keep an eye out in this instance whenever we try to rebuild it (which shouldn't be often)

An alternative would be to give the conda package a different, pep440 compliant version string that differs from the upstream number

marscher:
We could add a blacklist file for non pep440 compliant packages.

PR #468, opened 2016-03-26
added a blacklist for non pep440 compliant packages

marscher:


jchodera:
Cool, thanks!

Addresses #467

mpharrigan:
Thanks!

PR #469, opened 2016-03-28
Check if .pre_black_listed is available before trying to open it

rmcgibbo:
#468 broke conda-dev-recipes which didn't have a file called `.pre_black_listed`.


jchodera:
Thanks!

PR #471, opened 2016-03-29
Release 0.7.0 for openmoltools

andrrizzi:


jchodera:
`openmoltools` failed because of the `GFORTRAN_1.4` issue. Let me delete the problematic `ambermini` build and retrigger.

andrrizzi:
The `GFORTRAN_1.4` issues are gone, thanks! Looks like we still have some troubles. I'll first add `openmmtools` to the test requirements and see if we still have errors.

jchodera:
Shouldn't we use openmmtools-dev? Otherwise we bring in both openmm and openmm-dev

andrrizzi:
> Shouldn't we use openmmtools-dev? Otherwise we bring in both openmm and openmm-dev

Hmm, I'm not sure I understand. The [current conda recipe](Shouldn't we use openmmtools-dev? Otherwise we bring in both openmm and openmm-dev) uses `openmm` right now, not `openmm-dev`. Should we change that?

Also, it looks like now conda [is installing `ambermini 15.0.4-1`](https://travis-ci.org/omnia-md/conda-recipes/jobs/119289652#L3775) and we get some new [`GLIBC_2.7 not found` errors](https://travis-ci.org/omnia-md/conda-recipes/jobs/119289652#L4411). Is this maybe the `pandas` package (it's the only one that differs from the `choderalab/openmoltools` recipe)?

I've also noticed that even with previously successful builds on the `openmoltools` repo, travis was grabbing `ambermini 15.0.4-3`. Are there conflicts with other packages that prevent build 4 to be installed?

jchodera:
Sorry, I was confused!

jchodera:
> I've also noticed that even with previously successful builds on the openmoltools repo, travis was grabbing ambermini 15.0.4-3. Are there conflicts with other packages that prevent build 4 to be installed?

Build 4 has the `libgfortran` dependency, so it's possible this conflicts with the `gcc` dependency that brings in `GLIBC_2.7`. I can delete `ambermini 15.0.4-1` from anaconda cloud too---maybe that would fix things?

andrrizzi:
> I can delete ambermini 15.0.4-1 from anaconda cloud too---maybe that would fix things?

Maybe. If there is something wrong with 15.0.4-4 we'll probably get some errors on build, but at least we'll be able zero in on the problem. We also risk to remain with no functioning build for `15.0.4`.

jchodera:
> Maybe. If there is something wrong with 15.0.4-4 we'll probably get some errors on build, but at least we'll be able zero in on the problem. We also risk to remain with no functioning build for 15.0.4.

I think the problem might be a conflict between the `gcc` package and `libgfortran` package version. However, there's now a `libgfortran 3.0-0` package that I think we might be able to build `ambermini` against. Let me try that now.

jchodera:
Actually, nevermind. That doesn't seem to be it---`conda` happily lets me install `gcc` alongside `gfortran==1.0`

andrrizzi:
Now that builds 0 and 1 aren't there, `conda` does install `ambermini-15.0.4-4` and `libgfortran-1.0` for build, but it switches to `ambermini-14.0.1` and `libgfortran-3.0` (only with `python2`) when the test phase starts and `scipy-0.17.0-np110py27_2` is installed instead of `scipy-0.17.0-np110py27_1`.

Since the `libgfortran` dependency is brought in by `scipy`, should we try to install `scipy 0.16` instead of `0.17`?

andrrizzi:
It looks like we got rid of the weird `G*` errors! Although there are a bunch of test failing. Why weren't those tests failing in the main repo?

andrrizzi:
Oh, I guess we're using `openmm-dev` there?

andrrizzi:
tagging @swails that some time ago mentioned he had the same fortran problem with the pysander conda package

andrrizzi:
Setting `libgfortran 1.* # [linux and np == 19]` in the run requirements worked without having to fix `scipy` version. Thanks @mpharrigan for the solution!

We still have those force field problems though (presumably due to out-dated `OpenMM`?). @jchodera could you take a look?

jchodera:
Those errors are because the needed features are not in `openmm` but in `openmm-dev`.  I guess it's too early to cut an actual release.

andrrizzi:
I see. I'll revert the changes then.

jchodera:
Sorry for not catching that earlier.

I'd leave this open. 7.0rc1 is out and 7.0 is being released in just a few
days.

andrrizzi:
Good! Let's wait then.

andrrizzi:
@jchodera, could you retrigger the tests? I don't have the rights. Also, the recipe uses `openmm`, is that going to be a problem? Do we need to specify a version?

jchodera:
I think we need to pin `openmm` to `openmm ==7.0rc1` in the recipe.  Can you check that change in?

andrrizzi:
Ok!

andrrizzi:
Tests pass! I think we're ready to merge this.

jchodera:
Woo!  Thanks!

PR #473, opened 2016-03-29
[ambermini] Attempt to solve library compatibility issues

jchodera:


jchodera:
@andrrizzi: We're now seeing:
```
sqm: /anaconda/envs/_test/bin/../lib/libgfortran.so.3: version `GFORTRAN_1.4' not found (required by sqm)
```
That's very odd.

jchodera:
Seems relevant: https://github.com/ContinuumIO/anaconda-issues/issues/686

andrrizzi:
That link blames `scipy` too. I took a shot at installing `scipy-0.16.1` in #471 .

jchodera:
OK, it looks like ditching `libgfortran` for just `libgcc` seems to work, which is weird.

jchodera:
@andrrizzi : Do you think we should switch the dependency to `scipy-0.16.1` or should we give this current build a shot?

andrrizzi:
Yeah, if I remember correctly we added the `libgfortran-1.0` dependency because `scipy` was installing `libgfortran-3.0` and we wanted to make sure that the correct fortran compiler was installed with `ambermini`. Using `scipy-0.16.1` seems to work in #471 (although we have tests failing).

andrrizzi:
The `openmoltools` build without the dependency of `libgfortran-1.0` in `ambermini` was broken so I wouldn't bother trying it alone, but maybe we don't need it if we use `scipy-0.16.1`.

jchodera:
OK, merging this, since we don't have a `scipy` dependency in here.

andrrizzi:
Yep! I made a local test, `scipy-0.16.1` doesn't install `libgfortran` so we can in theory ditch the dependency if we want.

Is there a mechanism to say to a recipe what NOT to install?

PR #480, opened 2016-04-01
[openmm] 7.0rc1 build 1

jchodera:


jchodera:
Reworking this for #478

PR #481, opened 2016-04-01
[openmm] 7.0rc1 build 1

jchodera:


jchodera:
@peastman : Hopefully fixed!

peastman:
Great!

mpharrigan:
I don't think `git_url` and `git_rev` are appropriate here. It's still a zipfile, albeit one served by github

jchodera:
Whoops!  Thanks for catching that.  Not sure how that happened.

PR #483, opened 2016-04-03
[autoprotocol] Add recipe from conda skeleton

jchodera:


mpharrigan:
Looks like the license is bsd

https://github.com/autoprotocol/autoprotocol-python/blob/master/LICENSE.rst

PR #487, opened 2016-04-09
[WIP] setup conda packaging for meld-openmm-plugin

jlmaccal:
@jchodera fork as requested.


jchodera:
Awesome!  Let's see where this breaks.

jchodera:
You might want to more completely copy the [`openmm` recipe's `meta.yaml`](https://github.com/omnia-md/conda-recipes/blob/master/openmm/meta.yaml), but this is fine for a start.

jlmaccal:
Shit, it died, but for a stupid reason. Let me update and try again.

jchodera:
Whoops, we don't have your `eigen3` package in `omnia`. The [build error](https://travis-ci.org/omnia-md/conda-recipes/jobs/121982417#L266-L267):
```
Error:  Package missing in current linux-64 channels: 
  - eigen3
```

jchodera:
Oh, I see you got it already. (Hadn't refreshed.)

jchodera:
You need to add `--yes` so you have `conda install --yes eigen3`, but the usual way we handle this is to add the `eigen3` recipe too.

jlmaccal:
@jchodera @peastman Alright, sorry about that. It took me a while to figure out where conda was stashing the requirements. This is the same error as I was running into previously.

It fails while running:
```
/opt/rh/devtoolset-2/root/usr/bin/c++   -I/usr/include/nvidia/   CMakeFiles/TestCudaMeldForce.dir/TestCudaMeldForce.cpp.o  -o TestCudaMeldForce  -L/anaconda/envs/_build/lib  -L/anaconda/envs/_build/lib/plugins -rdynamic ../libMeldPluginCUDA.so ../../../libMeldPlugin.so /usr/local/cuda-7.5/lib64/libcudart_static.a -lpthread -lrt -ldl -lOpenMM -lOpenMMCUDA -Wl,-rpath,/anaconda/envs/_build/lib:/anaconda/envs/_build/lib/plugins:/anaconda/conda-bld/work/build/platforms/cuda:/anaconda/conda-bld/work/build
```

It is trying to build the executable for the unit tests. It is linking to openmm with `-lOpenMMCUDA`, but this seems to be giving a linker error as it cannot find `libcufft.so.7.5` for some reason.

jlmaccal:
I added the example plugin, which fails with the same issue.

@peastman Any ideas how to fix this? This hasn't been a problem before with manual builds.

jchodera:
Thanks!  Now @peastman and I have a reproducible way to debug this.  

No immediate ideas, but a bit of tinkering tomorrow might solve it.

peastman:
Just looking at the error in travis:

```
/opt/rh/devtoolset-2/root/usr/bin/c++   -I/usr/include/nvidia/   CMakeFiles/TestCudaExampleForce.dir/TestCudaExampleForce.cpp.o  -o TestCudaExampleForce  -L/anaconda/envs/_build/lib  -L/anaconda/envs/_build/lib/plugins -rdynamic ../libExamplePluginCUDA.so ../../../libExamplePlugin.so /usr/local/cuda-7.5/lib64/libcudart_static.a -lpthread -lrt -ldl -lOpenMM -lOpenMMCUDA -Wl,-rpath,/anaconda/envs/_build/lib:/anaconda/envs/_build/lib/plugins:/anaconda/conda-bld/work/build/platforms/cuda:/anaconda/conda-bld/work/build 
/opt/rh/devtoolset-2/root/usr/libexec/gcc/x86_64-CentOS-linux/4.8.2/ld: warning: libcufft.so.7.5, needed by /anaconda/envs/_build/lib/plugins/libOpenMMCUDA.so, not found (try using -rpath or -rpath-link)
```

Shouldn't there be a `-L/usr/local/cuda-7.5/lib64` in there?

jlmaccal:
@peastman Yes, agreed that's probably all that's needed to fix it.

My cmake-fu is virtually nonexistent, so I don't know why it's not being included. For what it's worth, I've never had a problem with building using `ccmake`. This problem is only cropping up now that I'm trying to use `cmake` and setting various flags. It's probably just a missing define or flag, but I don't know enough cmake to fix it.

peastman:
For the OpenMM build script, we specifically link against `${CUDA_cufft_LIBRARY}`.  You probably want something similar to that.

https://github.com/pandegroup/openmm/blob/master/platforms/cuda/sharedTarget/CMakeLists.txt#L16

jlmaccal:
Why is this necessary with `cmake` but not `ccmake`? I can build both packages manually just fine without changing `CMakeLists.txt`.

peastman:
There's no difference between `cmake` and `ccmake`.  The latter is just a UI for the former.

Just a guess: when you run it manually, do you have `/usr/local/cuda/lib64` in your library path?

jlmaccal:
Yes, it's in the library path, but it should be in the build scripts as well.

peastman:
Try specifically telling CMake about that library in the same way that OpenMM does.

jlmaccal:
@peastman I'll try that today, but the tests for the cuda platform don't link to `cufft` directly either:

https://github.com/pandegroup/openmm/blob/master/platforms/cuda/tests/CMakeLists.txt#L22

jlmaccal:
Oops... I wasn't setting `LD_LIBRARY_PATH` properly...

Now it's failing when building the python wrappers:
```
Scanning dependencies of target PythonInstall
[100%] Generating MeldPluginWrapper.cpp
/anaconda/envs/_build/bin/swig: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by /anaconda/envs/_build/bin/swig)
make[3]: *** [python/MeldPluginWrapper.cpp] Error 1
make[2]: *** [python/CMakeFiles/PythonInstall.dir/all] Error 2
make[1]: *** [python/CMakeFiles/PythonInstall.dir/rule] Error 2
make: *** [PythonInstall] Error 2
```

jchodera:
This is a `glibc` incompatibility problem, where it is saying the build system is using a `swig` version compiled with a newer `glibc` than is supported.  The `holy-build-box` only has the older `glibc` to make sure whatever it compiles is maximally compatible with older `linux` systems.

I think we usually get around this by using the `omnia`-provided `swig` to build:
https://anaconda.org/omnia/swig

Do you have `swig` as a `build` dependency in your conda recipe?

jchodera:
OK, you definitely have `swig ==3.0.7` in the recipe, so I'm super confused as to why this is happening.

jchodera:
One issue that may not be a problem: Your conda package and recipe directory name are different. I know that will cause problems down the road---they have to be the same for our build scripts to be happy.  I don't think this problem is related, though.

jlmaccal:
OK, so it looks like it's pulling `swig` from the channel where I grabbed `eigen3`, rather than from `omnia`.

jchodera:
Can you add the channels in opposite order, or add a recipe for `eigen3` to omnia?

jlmaccal:
Eventually I'll add `eigen3` to omnia, but for now I can just pull it from the channel directly, rather than adding the channel.

jchodera:
Here, let me just copy it to the `omnia` channel.

jchodera:
Boom: https://anaconda.org/omnia/eigen3/files

jlmaccal:
Closing this to resubmit a cleaned up version.

mpharrigan:
You can do that without opening a new pull request by rebasing and doing `git push --force`

PR #489, opened 2016-04-12
Add openmm-example-plugin

jlmaccal:
This adds the example plugin. 

While this doesn't add any useful end-functionality, I think it could be helpful to catch any changes that break plugins.

There is no windows build, as I'm unsure how to set that up correctly.

### DIFF
--- a/openmm-example-plugin/build.sh
+++ b/openmm-example-plugin/build.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+CMAKE_FLAGS="-DCMAKE_INSTALL_PREFIX=$PREFIX -DBUILD_TESTING=OFF"
+
+CMAKE_FLAGS+=" -DCMAKE_BUILD_TYPE=Release"
+
+if [[ "$OSTYPE" == "linux-gnu" ]]; then
+    CUDA_PATH="/usr/local/cuda-7.5"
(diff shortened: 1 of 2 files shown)


jchodera:
Looks like you stripped out all the `osx` stuff from the `openmm` recipe, but didn't have it skip `osx` builds in `meta.yaml`.

jlmaccal:
Yes, but it should go back in the recipe as windows and osx are supported by the example plugin.

jlmaccal:
@peastman @jchodera 

How do we currently build the conda packages for `openmm` on windows? I'm trying to figure out what to do with the example plugin, but it looks like apveyor is skipped for `openmm`. Should I just skip windows builds for the example plugin as well?

jchodera:
> Should I just skip windows builds for the example plugin as well?

I think so.  We just don't have a way to get OpenCL or CUDA on AppVeyor right now, though many people have asked for this capability (e.g. [here](http://help.appveyor.com/discussions/suggestions/682-install-intel-opencl-code-builder) and [here](http://help.appveyor.com/discussions/suggestions/806-install-amd-app-sdk-opencl)).

Currently, @peastman builds OpenMM `win` conda packages manually.

Eventually, we might be able to use Azure or a VM to do this, but we have zero expertise with `win` (and zero motivation to develop any).

jlmaccal:
OK, then this is good to merge for now. We can resist a windows recipe when the time comes.

PR #494, opened 2016-04-14
Add meld-plugin

jlmaccal:


jlmaccal:
@jchodera This should be ready to merge.

jchodera:
Woohoo!

PR #503, opened 2016-04-25
pyemma v2.1.2

marscher:


franknoe:
I don't think we want to release anything before resolving #782. Also #764 is not yet fixed.

marscher:
Well, we introduced a memory leak in the last version, which in my honest opinion legitimate a micro release.

franknoe:
Please see #782 - no releases before proper testing. Notebooks are failing with this version.

franknoe:
ok, let's do this:

1. check+merge #783
2. solve #764
3. update IPython notebook output on pyemma.org
4. if all looks OK, release.

PR #505, opened 2016-04-29
Add eigen3

jlmaccal:
@jchodera you previously copied the `eigen3` package to the omnia channel, but this was just the linux version. We might as well build our own for both mac and linux.


jlmaccal:
This should be good to merge, but you might want to remove the existing package from anaconda.

jchodera:
Done and done!

PR #509, opened 2016-05-03
Send packages to "rc" label by default. Support "--dev"

### LINKED ISSUE #488: Use labels correctly
http://docs.anaconda.org/using.html#UsingLabelsInTheDevelopmentCycle

Procedure for releasing a package where you want to be careful

1. bump version to what you want the final release to be called, e.g. "2.0"
2. upload to a label "rc" (release candidate)
3. when satisfied with the binary package, change the label to "main" *without* rebuilding

For (2), we can use the http://conda.pydata.org/docs/building/meta-yaml.html#extra-section extra section to figure out what label to upload to. 

We can also make everything go to the rc label first, but this might be overkill

mpharrigan:
Fixes #488

### DIFF
--- a/conda-build-all
+++ b/conda-build-all
@@ -69,6 +69,12 @@ def main():
         dest='force',
         help='Force a package upload regardless of errors',
     )
+    p.add_argument('--dev',
+       action='store_true',
+       dest='dev',
+       help=('Push package to the dev label. This requires '
+             '--force to be enabled.'),
+    )
     p.add_argument(
         '--no-test',
         action='store_true',
@@ -208,7 +214,8 @@ def list_package_contents(m, python, numpy):
             print('...')
             break
 
-def upload_to_binstar(m, python, numpy, username, max_retries=5, force=False):
+def upload_to_binstar(m, python, numpy, username, max_retries=5,
+                      force=False, dev=False):
     filename = get_bldpkg_path(m, python, numpy)
     if not os.path.exists(filename):
         print('ERROR: File does not exist: %s' % filename)
@@ -220,8 +227,15 @@ def upload_to_binstar(m, python, numpy, username, max_retries=5, force=False):
     cmd.extend(['upload', '--no-progress', '-u', username, filename])
     if force:
         cmd.extend(['--force'])
-    if not pre_black_listed(m) and is_prerelease(m) :
+    if not pre_black_listed(m) and is_prerelease(m):
         cmd.extend(['--label', 'pre'])
+    elif dev:
+        assert force, "Dev packages require forced uploads."
+        cmd.extend(['--label', 'dev'])
+    else:
+        # By default, packages go to the "rc" release candidate label
+        # Please test these binaries before switching them to the main label.
+        cmd.extend(['--label', 'rc'])
 
     err = None
     for i in range(max_retries):
@@ -304,7 +318,7 @@ def build_package(m, python, numpy, args):
     if args.upload is not None:
         if (retcode == 0):
             upload_to_binstar(m, python, numpy, username=args.upload,
-                              force=args.force)
+                              force=args.force, dev=args.dev)
         else:
             print('Package failed to build (return code %s); will not upload.' % str(retcode))
     sys.stdout.flush()


jchodera:
What about supporting a general `--label [labelname]` argument instead of separate `--dev` and `--pre` flags and a default behavior that sends things to `--rc`?  We can simply have the `omnia-md/conda-recipes` travis use `--label rc` (to always send packages to `rc`) and have individual packages use `--label dev`?

mpharrigan:
That makes sense, but I actually kindof like the limited flexibility by allowing only a curated set of labels. I can imagine a typo leading to a new label and confusion, or a new travis.yml omitting the label spec and having it to go the main label. There's also the precedent of the "pre" label, which uses logic in `conda-build-all` to determine the label. I included a little check here that people using the `dev` label are also using `--force`, which might prevent a headache or two

jchodera:
OK, let's run with this for now and change it later if necessary.

Passes travis, so feel free to merge whenever.  (Not sure if `win` will be a pain or not)

mpharrigan:
Great! I don't have merge permissions though, so I defer to you!

jchodera:
> I don't have merge permissions though

Now you do!

jchodera:
Whoops---could you also add some text to the `README.md`?  I totally forgot that needs to be updated too.

marscher:
Does this now mean that we have to manually move everything which is built to main label?

jchodera:
Yes!

marscher:
OMG! And how to move that stuff without using the website? This is really annoying.

jchodera:
We're happy to continue to tweak the process, but we thought this was important for quality control.

Is the website not working for you?

I believe there is a way to copy the package to main through the CLI:

http://docs.anaconda.org/using.html

See the section "Using labels in the development cycle"
 That example of copying the label presumably needs a channel and package specified, but I admit I was frustrated in getting that to work when I tried previously.

jchodera:
The goal was to be able to test packages before they get deployed to the main channel and go out to all users.

marscher:
Of course the website is working, but moving every file manually is a lot of hassle. I dont know how much time you want to spent on this, but are you really talking about testing every file for every platform for every version combination like manually before finally releasing? This would require a lot of man power in my opinion. 

In general it is a good idea, but why are you enforcing this kinda strict policy now to every package in the channel without really having a discussion about it?

rmcgibbo:
Does anyone know if there's a way to move files from `rc` to `main` through the CLI?

jchodera:
@rmcgibbo : According to [the docs](http://docs.anaconda.org/using.html#UsingLabelsInTheDevelopmentCycle), you're supposed to be able to use `anaconda label --copy` for this purpose, but I had trouble getting this to work on just one package.

jchodera:
> Of course the website is working, but moving every file manually is a lot of hassle. 

You can select multiple files and then move them all at once:
* Click on the upper-left checkbox
* Actions > Move > `main`

rmcgibbo:
I agree with @marscher that this new "feature" is a big hassle. I think it makes sense for some packages that have a well-developed RC process (OpenMM), but for most of the packages in this repo, it's just a hassle.

jchodera:
> In general it is a good idea, but why are you enforcing this kinda strict policy now to every package in the channel without really having a discussion about it?

I totally agree that this scheme may be overly strict. I was hoping we could implement something where each package could select which label their package gets uploaded to by putting something in `meta.yaml` or the recipe directory. I still think that may be a good way to go.

We *did* have a discussion, but (1) it was scattered in several threads, and (2) we didn't have a chance to talk about it in the google hangout meetings. I totally agree we could have been more inclusive in soliciting feedback.

The updates to `conda-build-all` actually broke the [`conda-dev-recipes`](https://github.com/omnia-md/conda-dev-recipes) and we haven't fixed that yet.  (We just need to add some way to send packages to `main` by default, by adding a command-line argument.)  We also figured we can use `conda-build-all` to build dev packages in the travis builds of individual repos to replace the scheme we use now (where we would use `packagename` version instead of `packagename-dev`), but this hasn't been implemented yet.

jchodera:
> I agree with @marscher that this new "feature" is a big hassle. I think it makes sense for some packages that have a well-developed RC process (OpenMM), but for most of the packages in this repo, it's just a hassle.

I agree.  How about we
* Have `conda-build-all` send packages to `main` by default
* Add an optional mechanism for packages to specify where they are uploaded
* Talk about whether we should adopt some QC process that involves sending to an `rc` channel first at the next omnia meeting

Does this sound reasonable?

Pinging @mpharrigan too.

jchodera:
@marscher : What should we do about the code that sends `X.YrcZ` packages to `pre`?  Should we leave that enabled?

marscher:
Actually we already have a mechanism to put packages in a "rc" channel,
but the channel is called "pre" and it depends on the version number (it
has to be recognized as pre release version according to versioning
standard pep440).

Of course a switch by package is also a nice idea!

marscher:
Actually I like the flag idea in meta.yaml more than an automatic decision based on version numbers. It gives more flexibility.

rmcgibbo:
In general, most of the packages here are developed in other repos, either by us, or by other developers. And we assume that the developers are following appropriate testing protocols there. So I don't think we honestly need that much QC in the package phase beyond "does the package import and perhaps pass a couple simple tests", which is exactly the `test` section of the `meta.yaml`.

jchodera:
> Actually we already have a mechanism to put packages in a "rc" channel, but the channel is called "pre" and it depends on the version number (it has to be recognized as pre release version according to versioning standard pep440).

This mechanism doesn't work for packages like OpenMM because it was deemed not "safe" enough. The package must be rebuilt to move from an `X.Yrc1` to an `X.Y.0` release, which leaves open the possibility that something breaks and produces problematic release packages.  

> Actually I like the flag idea in meta.yaml more than an automatic decision based on version numbers. It gives more flexibility.

I think this would be optimal.  If we could use something like
```YAML
build:
   label: rc
```
that would be ideal, but I'm not sure if this would break the parsing of the `meta.yaml` recipe that `conda_build.metadata` performs.

> In general, most of the packages here are developed in other repos, either by us, or by other developers. And we assume that the developers are following appropriate testing protocols there. So I don't think we honestly need that much QC in the package phase beyond "does the package import and perhaps pass a couple simple tests", which is exactly the test section of the meta.yaml.

This whole thing came about because @peastman insisted on a specific mechanism for releasing OpenMM that we did not support, and @mpharrigan wanted to try it for `mdtraj` as well.  I think you're right that it may not be necessary for *any* packages, though some might opt to use it.

marscher:
http://conda.pydata.org/docs/building/meta-yaml.html#extra-section

jchodera:
OK, how about
```YAML
extra:
  label: rc # defaults to `main`
```
Would that work?

marscher:
Fine for me, if the default behavior is main.

marscher:
Until this will be implemented I strongly encourage to revert the changes. It only causes problems for every project other than OpenMM!

jchodera:
It's going to be a mess to just revert the changes, since we've made several bugfixes to the scripts since then.

Let's see if we can just easily implement this in one shot, changing OpenMM and MDTraj as the only recipes to use the new scheme.

marscher:
I guess it only affects conda-build-all?

jchodera:
Yes, but `conda-build-all` has had several sets of changes merged in since then.  

Do you want to take a stab at implementing this feature, @marscher, or should I?

marscher:
I can do it.

rmcgibbo:
I've already moved over most of the new packages to main/, eg from the numpy 1.11

jchodera:
> I've already moved over most of the new packages to main/, eg from the numpy 1.11

Yikes. I totally spaced and hadn't realized this would push all the 1.11 packages to `rc`---so sorry about that!

jchodera:
And thank you!

marscher:
What do you think about having an explicit list in extras section like this:

```
extras:
  channel_labels:
    - pre
    - dev
    - main # this can be implicit

```

and then add this to the upload call? However we need to ensure, that
all these labels are being checked for prior uploading. Since we could
add more or less arbitrary labels in the recipes, we should standardize
them somehow.

jchodera:
Are you thinking that each recipe would specify two things?
1. the labels to be checked to see if the package needs to be uploaded
2. the label to which the package should actually be uploaded

I'd actually favor standardizing `main`, `dev`, and `pre` (I actually like `rc` or `preview` or `beta` or something more descriptive than `pre`) for all of `omnia`, and people can choose to use the ones they want by controlling which label the package is uploaded with.

marscher:
On 14.05.2016 23:27, John Chodera wrote:
> Are you thinking that each recipe would specify two things?
> 1. the labels to be checked to see if the package needs to be uploaded
> 2. the label to which the package should actually be uploaded
Currently the check is performed by adding an argument to
conda-build-all, but we should then extract it from the recipe.

> 
> I'd actually favor standardizing |main|, |dev|, and |pre| (I actually
> like |rc| or |preview| or |beta| or something more descriptive than
> |pre|) for all of |omnia|, and people can choose to use the ones they
> want by controlling which label the package is uploaded with.
I have no strong opinion about the name. We can use rc, if you think
that is more descriptive.

jchodera:
> Currently the check is performed by adding an argument to conda-build-all, but we should then extract it from the recipe. 

If we standardize the choices ('rc', 'dev', 'main') then we can keep this at the `conda-build-all` argument level.  I don't think I see a strong need to add the flexibility to let packages go crazy with labels, but I don't see it causing too many problems either if you wanted to generalize things so that each recipe specified something like
```YAML
extra:
  labels:
    upload: rc
    check:
    - rc
    - dev
    - main
```

marscher:
> If we standardize the choices ('rc', 'dev', 'main') then we can keep
> this at the |conda-build-all| argument level. I don't think I see a
> strong need to add the flexibility to let packages go crazy with labels,
> but I don't see it causing too many problems either if you wanted to
> generalize things so that each recipe specified something like
Yes we should keep it simple. I've seen you've added another flag to
force upload to the dev label. Do you want to keep this?
> 
> extra:
>   labels:
>     upload: rc
>     check:
>     - rc
>     - dev
>     - main
These three labels sound reasonable.

I think we would then only need to specify the upload label, since the
check is performed from level of invoking conda-build-all.

jchodera:
> Yes we should keep it simple. I've seen you've added another flag to force upload to the dev label. Do you want to keep this?

@mpharrigan added this with the idea that we could then use `conda-build-all` in our project `.travis.yml` to have travis build and upload the package (version 0.0.0) to a special `dev` channel. We would no longer need that with the `extras:` proposal here, so you can ditch that.

jchodera:
> I think we would then only need to specify the upload label, since the check is performed from level of invoking conda-build-all.

How about having these labels checked by default (`'main`, `dev`, `rc`) by `conda-build-all` if no `--check-against` flag is specified?

PR #515, opened 2016-05-03
Check against omnia/label/rc

### LINKED ISSUE #488: Use labels correctly

mpharrigan:
fixes #488


jchodera:
Restarted the `osx` build.

PR #490, opened 2016-04-12
Build numpy 1.11 packages

mpharrigan:

PR #517, opened 2016-05-10
Add numpy 1.11 support to conda-build-all defaults

jchodera:
`anaconda` has added numpy 1.11 support to many packages, so it's time we add support too.

### DIFF
--- a/conda-build-all
+++ b/conda-build-all
@@ -91,7 +91,7 @@ def main():
         '--numpy',
         help="Set the NumPy version used by conda build",
         metavar="NUMPY_VER",
-        default='19,110',
+        default='19,110,111',
     )
     p.add_argument(
         '-v', '--verbose',


mpharrigan:
cc #490

mpharrigan:
In particular, we should also *stop* building numpy 1.9 packages

jchodera:
Whoops, hadn't realized you were working on that!  

What is the status of the other PR?  Can you deprecate 1.9 and see if things pass?

jchodera:
@mpharrigan Any thoughts on this?  Can you comment on why the other PR (#490) is stalled?

mpharrigan:
Some of the recipes fail to build, as is the case here

jchodera:
None of the recipes here failed to build---travis just timed out because too many recipes tried to build.  

Or do you mean you tried locally and some failed?  Which ones?

mpharrigan:
```

[conda-build-all] Error: one or more packages failed to build (linux-64)

  bhmm-0.6.1-np111py27_0.tar.bz2

  pyemma-2.1.1-np111py27_0.tar.bz2
```

mpharrigan:
Travis linux and appveyor failed to build bhmm. osx is the only one that timed out

jchodera:
Can we skip those on numpy 1.11?  Does `meta.yaml` have syntax to do that?

mpharrigan:
The failure looks like it's with an actual test. Correct me if I'm wrong, but isn't the general scheme to have the full suite of tests run with CI and these recipes only have cursory tests to make sure the package was built correctly. [thorough testing can be performed with the `rc` binary]. Can we remove the nosetests from the recipe?

jchodera:
> Correct me if I'm wrong, but isn't the general scheme to have the full suite of tests run with CI and these recipes only have cursory tests to make sure the package was built correctly. 

100% correct!

PR #518, opened 2016-05-13
[yank] 0.8.0 release (redux)

jchodera:


jchodera:
This is very weird.  I'm not sure why it's not trying to build `yank`.

jchodera:
@mpharrigan: Any ideas? I don't understand why it isn't trying to build a new YANK release here.  Is there some way we can figure out why it is skipping the recipe?

jchodera:
Figured it out.  The `skip` constraint processing wasn't working as expected.

I have now added more debug info to `conda-build-all` so we can more easily tell why some packages aren't being scheduled for builds.

PR #522, opened 2016-05-14
Specify default upload label

marscher:
[conda-build-all] by default upload package to main. Recipes can specify label
    
    A meta.yaml can specify the channel to upload to within the extras section:
    
    extras:
      upload: rc
    
    Then all builds of this package will be uploaded specified channel only.
    
    Three default channels are now defined:
    1. main
    2. dev (development builds)
    3. rc (release candidates)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -65,5 +65,5 @@ build: false
 
 test_script:
   - ps: if (($env:APPVEYOR_PULL_REQUEST_NUMBER -eq $null) -and ($env:APPVEYOR_REPO_BRANCH -eq "master")) { $env:UPLOAD = "--upload omnia" } else { $env:UPLOAD = " " }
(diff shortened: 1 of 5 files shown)


marscher:
@jchodera I guess this is what we both had in mind?

jchodera:
This is great!

jchodera:
The travis output says that it is searching for the wrong channels, however:
```
channel_urls: ('omnia', 'omnia/label/pre', 'omnia/label/rc')
```
Any idea why it's not looking for `dev` instead of `pre`?

marscher:
@jchodera I've resolved this issue you reported. The labels were given explicitly by invocation of conda-build-all, so it has overridden the defaults.

jchodera:
Awesome, thanks!

jchodera:
Will merge this, but tagging @mpharrigan that you'll need to add this label to the `mdtraj` recipe!

mpharrigan:
Noted. Thanks guys!

PR #525, opened 2016-05-15
added python no_arch for nglview

marscher:


jchodera:
Do we need to increment the build number too, or did this not build successfully before?

marscher:
I think if one switches from an arch to a no arch package, one does not need to increment the build number to trigger a new build.

jchodera:
But won't the old arch-based packages still be on anaconda cloud and have the same build number? Or would you manually delete them?

PR #527, opened 2016-05-17
[conda-build-all] pre-releases go to rc now.

marscher:
see discussion in #488 

Since we only want to use main, dev, rc for now, automatically detected development versions (eg. alpha, beta, rc1 and so on) will be moved to the dev label. This used to be pre before. So we should locate and move files in "pre" and move them to dev in the next step.

@jchodera please review.


mpharrigan:
The title of this should be "go to dev now", right?

marscher:
indeed, thank you.

PR #530, opened 2016-05-23
Test graphviz 2.38.0-3

nividic:
Checking if the test part is passed and its output


rmcgibbo:
you probably want to increment the `build/number` field if you want to force a rebuild.

nividic:
Thanks I didn't know that. Effectively the tests are passed on both linux and osx.

PR #531, opened 2016-05-25
[openmm 7.0.1] build based on git hash 5e86c4f

jchodera:
See https://github.com/pandegroup/openmm/issues/1489

### DIFF
--- a/openmm/meta.yaml
+++ b/openmm/meta.yaml
@@ -1,13 +1,13 @@
 package:
   name: openmm
-  version: 7.0rc1
+  version: 7.0.1
 
 source:
   git_url: https://github.com/pandegroup/openmm.git
-  git_rev: f433936
+  git_rev: 5e86c4f
 
 build:
-  number: 1
+  number: 0
   skip:
     # don't build on Appveyor (no OpenCL / CUDA)
     - [ {{ os.environ.get('APPVEYOR', 'False') == 'True' }} ]


jchodera:
This should go to the `rc` label.

jchodera:
For some reason, OpenMM 7.0.1 didn't get built upon merge.

@marscher: Any idea why?

rmcgibbo:
I would guess because "7.0rc1" sorts in front of "7.0.1".

jchodera:
@marscher: Did you run into this problem too when you used the old pre label? When you moved it to main, did you have trouble with X.Yrc1 sorting ahead of X.Y.Z?

I suppose we could add some feature to conda-build-all to let us force the build of any package that isn't already in anaconda cloud, rather than just packages that are more up to date than what is on anaconda cloud? If the sorting problem is real, we would have to figure out how to replace OpenMM 7.0rc1 safely before moving on, though.

marscher:
Am 25.05.2016 um 07:29 schrieb Robert T. McGibbon:
> I would guess because "7.0rc1" sorts in front of "7.0.1".
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub <https://github.com/omnia-md/conda-recipes/pull/531#issuecomment-221476755>
>
Well I thought the behaviour is like the following: if the package is not found in the specified conda channels, a build will be triggered, no matter if it is a 
old version or a recent one? Of course we only have the "current" git head, which points to the desired version to build.

jchodera:
> Well I thought the behaviour is like the following: if the package is not found in the specified conda channels, a build will be triggered, no matter if it is a old version or a recent one?

The recipe just added is for `openmm-7.0.1`, but it does not seem to be building because `openmm-7.0rc1` is present.

I'll open a new pull request to get some more debug information about what is going on.

jchodera:
cc https://github.com/omnia-md/conda-recipes/pull/533

PR #540, opened 2016-06-11
Add SolvationToolkit recipe

davidlmobley:
SolvationToolkit allows easy setup of simulation inputs for arbitrary mixtures of small organic compounds. This adds a recipe for it.

This is my first crack at adding a new recipe for one of my own packages, so likely I'll have some issues at first. I've modified recipes before, though, so hopefully that means I haven't botched this too badly.


davidlmobley:
Ahh, right. Currently my tests require OpenEye tools to run, which is fine as long as I'm running them on MY Travis-CI, but won't work on here. Hence this is failing on Travis.

I knew there had to be some good reason to have the tests skip if OpenEye tools aren't available... 

I'll fix this soon and try again.

jchodera:
> I knew there had to be some good reason to have the tests skip if OpenEye tools aren't available...

Yep.

We've generally been excluding all tests (except a simple `import packagename`) from the official conda recipes included here, since otherwise builds+tests take forever.

davidlmobley:
@jchodera : This is failing because of something relating to a docker image of an omnia build box: 

```
 $ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then docker pull jchodera/omnia-build-box; fi
Using default tag: latest
latest: Pulling from jchodera/omnia-build-box
nia-build-box
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

Ideas?

jchodera:
This is a stochastic failure due to high network traffic

davidlmobley:
@rmcgibbo - do you have any advice on this? It seems to be timing out during the testing and I don't really have enough experience with this to troubleshoot much yet. I basically just copied the recipe from openmoltools (where John wrote it) and adjusted it for this, so I don't THINK it's a problem with the recipe.

rmcgibbo:
There is a big bug that's causing every single package to be scheduled for rebuild.

davidlmobley:
@rmcgibbo @mpharrigan - any chance you could help me understand what's going wrong with Travis and AppVeyor here? The errors I'm seeing don't seem to obviously have to do with my recipe, but maybe I've somehow asked for conflicting versions of things or for builds I shouldn't have... 

For AppVeyor I'm seeing things like this:

```
ERROR: File does not exist: C:\Miniconda\conda-bld\win-32\msmbuilder-3.4.0-np111py27_0.tar.bz2
Removing old build environment
BUILD START: pyemma-2.2-np111py27_0
Fetching package metadata ............. 
Solving package specifications: .Using Anaconda Cloud api site https://api.anaconda.org
... 
Error: The following specifications were found to be in conflict:
  - numpy 1.11*
  - python 2.7*
```

and for Travis-CI I'm getting a failure on OS X for Python 3.4 only. I'm seeing something like this:

```
TEST START: solvationtoolkit-0.4.1-py34_1
Using Anaconda Cloud api site https://api.anaconda.org
Fetching package metadata ...........
Solving package specifications: .
Error: Package missing in current osx-64 channels: 
  - solvationtoolkit 0.4.1 py34_1
ERROR: File does not exist: /Users/travis/anaconda/conda-bld/osx-64/solvationtoolkit-0.4.1-py34_1.tar.bz2
Removing old build environment
```

which is odd, because it seems inconsistent with what I'm seeing for the python2.7 and python2.5 builds (naively I was not expecting to see it looking for a build file yet at this point since this is a section where it looks like it's trying to start to build, and in the Python2.7 and Python3.5 sections it doesn't seem to be looking for a build file).

I'm just trying to sort out where I've gone wrong... 

Apologies for all of the "newb" questions here.

rmcgibbo:
The appveyor issue is not specific to this PR. If you look in the log for other recent PRs, the same issue is there. Some mdtraj/pyemma dependency is missing for py27np110. Not a big deal.

idk exactly what the py34 osx issue is though.

davidlmobley:
@rmcgibbo - does this mean we should be bypassing py27np110 somehow? Naively it seems like the point of tests is to only fail when something is wrong that we want to fix.

rmcgibbo:
You can add skips if you want. There's a trad eoff either way.

- On one hand: I'm just guessing, but the underling issue might be something like that Continuum hasn't uploaded a win32 py27np110 package for pytables. If we add a skip that prevents Appveyor from trying to build msmbuilder/pyemma for that configuration, and Continuum happens to remedy the issue tomorrow by uploading such a package, then we won't notice it and msmbuilder/pyemma will probably never get built for that configuration.
- On the other hand: if the build is red, you have to kind of read through the log file and understand the root causes of a failure when filing a PR, and know whether you should be merging even you see a red X.

Honestly, there's almost always something wrong here that's beyond our control such that it's very common for the PR tests to fail, even when the PR should be merged. So I think it's generally pretty necessary to read the log files and merge even in the face of a red X.

peastman:
We dropped support for mdtraj running under Python 2.7 on Windows, because no one was willing to  maintain it.  See https://github.com/mdtraj/mdtraj/pull/1125.

rmcgibbo:
FWIW, it looks from the log like the windows build for msmbuilder/mdtraj is failing for every version of python with np111, so it's not just a Python 2.7 thing.

mpharrigan:
> We dropped support for mdtraj running under Python 2.7 on Windows, because no one was willing to maintain it. See mdtraj/mdtraj#1125.

We haven't made a release that drops that support yet though

davidlmobley:
@jchodera - is this read to merge? I'm still not used to looking at the output of the tests on conda recipes. I tried to mirror what you're doing in `smarty` with the meta.yaml; it's passing some of travis's tests, and none of appveyor, but it looks like the last PR you merged behaved somewhat similarly (no success on appveyor).

solvationtoolkit/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - openmm
+    - pytables
+  run:
+    - python
+    - setuptools
+    - numpy
+    - numpydoc
+    - scipy
+    - ambermini
+    - parmed-dev
```
jchodera: Change this to `parmed`

solvationtoolkit/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - numpy
+    - numpydoc
+    - scipy
+    - ambermini
+    - parmed-dev
```
jchodera: Change this to `parmed`

solvationtoolkit/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - netcdf4
+    - pandas
+    - openmm
+    - pytables
+
+test:
+  requires:
+    - nose
+    - openmmtools
+    - openmoltools
```
jchodera: Don't need `openmoltools` here since it is already specified in `run`.

solvationtoolkit/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+#  imports:
```
jchodera: Can you at least enable the `imports` test to make sure this was built properly?

solvationtoolkit/meta.yaml
davidlmobley: Fixed.

solvationtoolkit/meta.yaml
davidlmobley: Fixed.

solvationtoolkit/meta.yaml
davidlmobley: Fixed.

davidlmobley:
@jchodera - comments addressed (except for the `import` test question above) so it may be ready to merge depending on your answer to that.

solvationtoolkit/meta.yaml
jchodera: Ah, good point---we can't install a pip dependency there.

PR #547, opened 2016-06-23
update osprey

cxhernandez:


cxhernandez:
Appveyor seems to be failing all across the board even though windows builds have been skipped. Is this still okay to merge?

PR #552, opened 2016-07-10
Release msmb_data v2.0

mpharrigan:
This package puts a couple hundred MB of example molecular dynamics data into your python prefix's `share/msmb_data` directory. We provide an API to fetch this data in `msmbuilder`. Other packages can use this data by loading the files

```python
import sys
"{prefix}/share/msmb_data/{dataset}/...".format(prefix=sys.prefix, dataset='fs_peptide')
```

### DIFF
--- a/msmb_data/bld.bat
+++ b/msmb_data/bld.bat
@@ -1,8 +1,3 @@
-"%PYTHON%" setup.py install
+mkdir "%PREFIX%\share\"
+xcopy msmb_data "%PREFIX%\share\msmb_data" /e /i /y
 if errorlevel 1 exit 1
-
-:: Add more build steps here, if they are necessary.
-
-:: See
-:: http://docs.continuum.io/conda/build.html
-:: for a list of environment variables that are set during the build process.
--- a/msmb_data/build.sh
+++ b/msmb_data/build.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
-
-$PYTHON setup.py install
+mkdir -p $PREFIX/share/
+cp -r msmb_data $PREFIX/share/msmb_data
--- a/msmb_data/meta.yaml
+++ b/msmb_data/meta.yaml
@@ -1,42 +1,17 @@
 package:
   name: msmb_data
-  version: "1.0.0"
+  version: "2.0.0"
 
 source:
   git_url: https://github.com/msmbuilder/msmb_data.git
-  git_tag: 043676571130dab6c3fcca60d49a8d5f4dc99c59
+  git_tag: v2.0
 
 extra:
   maintainers:
    - cxhernandez
-
-requirements:
-  build:
-    - python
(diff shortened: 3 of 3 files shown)


jchodera:
We have to be a little bit careful with adding packages that are a few hundred MB in size, since we have limited org storage. We're only using 15 / 100 GB now, but if everybody starts doing this as well, and if we update the files a few times, we might run into storage issues.
image

franknoe:
Is it needed to put mass data in a repo? I guess the history will also 
explode if you change files in future version.
What about putting mass data on an otherwise accessible download server 
and just having the download script in the repo?

Am 10/07/16 um 22:15 schrieb John Chodera:
>
> We have to be a little bit careful with adding packages that are a few 
> hundred MB in size, since we have limited org storage. We're only 
> using 15 / 100 GB now, but if everybody starts doing this as well, and 
> if we update the files a few times, we might run into storage issues.
> image 
> <https://cloud.githubusercontent.com/assets/3656088/16715839/750df8c4-46b9-11e6-8a04-3999e4351155.png>
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub 
> <https://github.com/omnia-md/conda-recipes/pull/552#issuecomment-231608472>, 
> or mute the thread 
> <https://github.com/notifications/unsubscribe/AGMeQub4fyjWGW4VxAiYgOnpGhlotzUEks5qUVLIgaJpZM4JI5f0>.
>


-- 

----------------------------------------------
Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de
Mail:  Arnimallee 6, 14195 Berlin, Germany
----------------------------------------------

jchodera:
> Is it needed to put mass data in a repo? I guess the history will also explode if you change files in future version. What about putting mass data on an otherwise accessible download server and just having the download script in the repo?

I think this is the best option: The code could download and cache the data the first time you run/import it.

We could use [Amazon S3](http://aws.amazon.com/s3/) for this, though this may cost money. Sharing S3 buckets might be hard---I'm not sure how sharing credentials works---so it might be best for each tool to maintain their own.

Alternatively, we could have a server hosted at an academic institution where bandwidth/hosting is free, but maintenance and hardware would cost money, and the durability might be questionable.

mpharrigan:
That's good to know about the limits. There was a bit of discussion on this package's repository about whether it could include a convenience python api for accessing the data. While that sounds nice, it means that we'd need a conda package for each python version! Since this doesn't depend on python, there's only one build per version (maybe one for each platform). 

If space becomes a problem, we can remove older versions of this package. We anticipate that this will remain pretty stable and not have many releases. The git history is a separate issue; it doesn't affect the conda package or our anaconda usage

PR #556, opened 2016-07-13
[openmoltools] Fix openmoltools 0.7.1 openmm dependency

jchodera:


jchodera:
I had accidentally kept the `openmm ==7.0rc1` runtime dependency in the previous attempt. This fixes that, so that both `build` and `run` require `openmm >=7.0.1`.

PR #565, opened 2016-08-04
[openpathsampling] 0.9.0

jchodera:
Initial tests. DO NOT MERGE!


marscher:
you should not use ssh+git in git_url, but just the https protocol.

jchodera:
@dwhswenson and @jhprinz : You don't expect python 3.x support, right? Looks like there's non-python 3.x compatible code in `openpathsampling`.

jchodera:
I'm seeing the same failure the `master` is failing with:
```
===== testing package: openpathsampling-0.9.0-py27_0 =====
import: 'openpathsampling'
/anaconda/envs/_test/lib/python2.7/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment.
  warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')
  File "/anaconda/conda-bld/test-tmp_dir/run_test.py", line 26, in <module>
    import openpathsampling
  File "/anaconda/envs/_test/lib/python2.7/site-packages/openpathsampling-0.9.0-py2.7.egg/openpathsampling/__init__.py", line 33, in <module>
    from analysis.move_scheme import (
  File "/anaconda/envs/_test/lib/python2.7/site-packages/openpathsampling-0.9.0-py2.7.egg/openpathsampling/analysis/__init__.py", line 1, in <module>
    from histogram import Histogram, SparseHistogram, HistogramPlotter2D
  File "/anaconda/envs/_test/lib/python2.7/site-packages/openpathsampling-0.9.0-py2.7.egg/openpathsampling/analysis/histogram.py", line 4, in <module>
    import matplotlib.pyplot as plt
  File "/anaconda/envs/_test/lib/python2.7/site-packages/matplotlib/pyplot.py", line 114, in <module>
    _backend_mod, new_figure_manager, draw_if_interactive, _show = pylab_setup()
  File "/anaconda/envs/_test/lib/python2.7/site-packages/matplotlib/backends/__init__.py", line 32, in pylab_setup
    globals(),locals(),[backend_name],0)
  File "/anaconda/envs/_test/lib/python2.7/site-packages/matplotlib/backends/backend_qt4agg.py", line 18, in <module>
    from .backend_qt5agg import FigureCanvasQTAggBase as _FigureCanvasQTAggBase
  File "/anaconda/envs/_test/lib/python2.7/site-packages/matplotlib/backends/backend_qt5agg.py", line 16, in <module>
    from .backend_qt5 import QtCore
  File "/anaconda/envs/_test/lib/python2.7/site-packages/matplotlib/backends/backend_qt5.py", line 31, in <module>
    from .qt_compat import QtCore, QtGui, QtWidgets, _getSaveFileName, __version__
  File "/anaconda/envs/_test/lib/python2.7/site-packages/matplotlib/backends/qt_compat.py", line 137, in <module>
    from PyQt4 import QtCore, QtGui
ImportError: No module named PyQt4
TESTS FAILED: openpathsampling-0.9.0-py27_0
ERROR: File does not exist: /anaconda/conda-bld/linux-64/openpathsampling-0.9.0-py27_0.tar.bz2
[conda-build-all] Error: one or more packages failed to build (linux-64)
  openpathsampling-0.9.0-py27_0.tar.bz2
```

dwhswenson:
> You don't expect python 3.x support, right?

Correct. OPS `1.0` is not planned to be python 3 compatible. I keep joking that we're saving that for OPS version `1.(python)3`.

jchodera:
cc: https://github.com/openpathsampling/openpathsampling/pull/584

jchodera:
Whoops. We don't support win either, do we @dwhswenson?

jhprinz:
Well, the package is exclusively Python Code and if all requirements exist for windows we should, right? All I can think of is that we did not use consistant file naming, which results in problems.

jchodera:
Oh! Looks like the problems may be openmmtools and python-coveralls.

dwhswenson:
As @jhprinz says, we're pure Python. However, tests might have some hard-coded `/` as path separators (we tried not to, but some might have slipped).

Looking at the Appveyor build, it appears that the problem is that `python-coveralls` doesn't exist in the Windows channels. That's used to send coverage info to coveralls.io, but it's fine if we only use do it from Travis. Is there a way to have that only in the conda test environment if we're on a non-Windows system?

jhprinz:
Yes, there is. You can add a filter to force certain lines only executed on specific platforms like this:
```
test:
  requires:
    - nose
    - nose-timer
    - python-coveralls  # [not win]
    - msmbuilder
    - pyemma
    - ipynbtest
```

jhprinz:
Actually I think coveralls should not be part of the test, we add the python coveralls to the test environment by hand. We only used this with the old conda-build.

openpathsampling/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - openmmtools
+    - pyyaml
+    - mdtraj
+    - svgwrite
+    - networkx
+    - pyqt 4*    
+    - matplotlib
+
+test:
+  requires:
```
jhprinz: I think you can remove the requires section. It is only needed for the import test and you should not need any additional packages for this. We used this before because we ran the nose tests, etc in the same environment. We now do it by hand.

jchodera: Thanks, will do!

jchodera:
I'm fixing the `openmmtools` `win` dependency now.

cc: https://github.com/omnia-md/conda-recipes/pull/589

jchodera:
The `openmmtools` `win` package has been uploaded, so testing this now.

jchodera:
Oh, well, there's not `win` support for python 2.7, so that was futile.

jhprinz:
Means, you can use OPS on windows but not with openmmtools. Seems we should get py3 support as soon as possible...

jchodera:
> Seems we should get py3 support as soon as possible...

Yup!

PR #576, opened 2016-08-13
[progress_reporter] build univeral wheel to avoid conda build bug

marscher:


jchodera:
Huh. Was the bug affecting windows builds only? Is there a link to the bug issue?

marscher:
slash vs. backslash mistake on my side... 

Bug report is here: conda/conda-build#1209

jchodera:
> Bug report is here: conda/conda-build#1209

Awesome, thanks!

marscher:
cmd.exe never supported wildcard expansion :+1:

PR #577, opened 2016-08-30
add msme

cxhernandez:
Adding a recipe for [msmexplorer/msmexplorer](https://github.com/msmexplorer/msmexplorer) (re: https://github.com/msmexplorer/msmexplorer/issues/26)

### DIFF
--- a/msmexplorer/bld.bat
+++ b/msmexplorer/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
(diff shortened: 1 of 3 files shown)


cxhernandez:
Looks like it I would need to either add a recipe for [dfm/corner.py](https://github.com/dfm/corner.py) or add the astropy channel to the conda build environment. Any preference in how to proceed, @jchodera @mpharrigan?

marscher:
If we do not add the corner.py recipe here, every user would have to add the astropy channel too in order to resolve the dependency. So it is a clear no go IMHO.

cxhernandez:
@marscher: Ah, okay. I'll try to add a new recipe for it then.

PR #582, opened 2016-09-14
Removing old keras/theano from Omnia

rbharath:
This PR removes the Omnia version of `keras`/`theano`. The Omnia versions of these packages were quite old, and I believe that `deepchem` was the only package that used these old versions (I added the `keras`/`theano` folders to Omnia previously for `deepchem` support). Since deepchem/deepchem#227 upgraded `deepchem` to use the latest `keras`/`theano`, I think we can go ahead and get rid of them.


rmcgibbo:
+1 from me.

jchodera:
I think this raises an important point: Should we also purge the conda packages in the `omnia` channel when we do this, or keep them there for older packages that may depend on them for reproducibility?

I think if we keep them, we'd still pick up later versions if we have the appropriate channels added. If we remove them, we make it impossible (or very hard) to reproduce work done with older packages that depend on them.

So I'd vote we keep the uploaded packages, though remove this recipe.

rmcgibbo:
I would generally be fine for purging conda packages in the omnia channel at the discretion of the maintainers. I think it's appropriate, for example, when the package is widely available elsewhere and isn't central to anyone's published science.

For example, omnia/conda-recipes (for some god-aweful reason) includes a recipe for [`m4`](http://www.gnu.org/software/m4/m4.html). Hopefully whatever required this can now depend on something in the main channels or conda-forge. If that's the case, I would feel a little silly keeping the m4 packages around.

jchodera:
How can we be sure that removing m4 or some other package won't break
existing packages or earlier versions thereof? If the dependency wasn't
pinned to a specific version, adding another conda channel MAY allow that
package to be correctly installed, but if the version was pinned to some
version no longer available or one that is somehow noninteroperable, we've
just destroyed the ability to reproduce the work with the older package.

I think we are committed to not removing old packages unless we are really
sure it won't break any released versions of omnia tools. We can certainly
remove recipes, but the potential for causing problems by removing packages
seems too great.

rbharath:
@jchodera @rmcgibbo Just wanted to give this a poke. It sounds like the issue of package removal is orthogonal to this PR. Can we go ahead and merge this one in?

jchodera:
Sure, but we still have to figure out what to do with the old package.
Are you using a constraint to force picking up a later version?

rbharath:
@jchodera I removed the dependency on omnia for `keras/theano` entirely and install directly from pip for both (changed `deepchem` install directions and travis-CI). In this case, I think we should be good to remove old package. I don't believe there are any other users left.

rbharath:
@jchodera PS, thanks for quick merge!

PR #583, opened 2016-09-15
update msme

cxhernandez:
msmexplorer 0.2.0 release

PR #584, opened 2016-09-15
Update schema to 0.6.2, enable py3k builds

Lnaden:
In part for the python 3 pledge, and to enable apps which depend on this to make their own py3k builds.

### DIFF
--- a/schema/meta.yaml
+++ b/schema/meta.yaml
@@ -1,15 +1,11 @@
 package:
   name: schema
-  version: !!str 0.5.0
+  version: !!str 0.6.2
 
 source:
-  fn: schema-0.5.0.tar.gz
-  url: https://pypi.python.org/packages/4b/eb/fadd6bc4967fded31495ea70aebd3bdc90b83e3944d08bfd841039062f83/schema-0.5.0.tar.gz
-  md5: cee375ebfa54f1c4a218438b3b1c6803
-
-build:
-  skip:
-    - [py3k]
+  fn: schema-0.6.2.tar.gz
+  url: https://pypi.python.org/packages/9b/3c/5924163d693f156684e1502df0b408512e18c70d7fa80244496fafdb5988/schema-0.6.2.tar.gz
+  mdf: ac1b359c31e9b37fbe9333902f6bdf76
 
 requirements:
   build:
--- a/schema/meta.yaml
+++ b/schema/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: schema
-  version: !!str 0.6.2
+  version: 0.6.2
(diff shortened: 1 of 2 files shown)


andrrizzi:
Not sure what's going on with the build, but it's probably not related to your changes as PR #583 seems to have the same problem. Still investigating.

andrrizzi:
[This is the script](https://github.com/omnia-md/conda-recipes/blob/master/conda-build-all#L39) that causes the error. It looks like for some reason Travis can't find basic Python conda packages.

Lnaden:
I think it might be either [first or last line](https://github.com/omnia-md/conda-recipes/blob/master/conda-build-all#L35-L37) of this block which is causing the `try:` block to fail. On my local box, `conda_build.config` does not exist and `.short_build_prefix` is not a property. A bit trickier to figure out if that's whats failing on the docker image.

schema/meta.yaml
```diff
@@ -1,15 +1,11 @@
 package:
   name: schema
-  version: !!str 0.5.0
+  version: !!str 0.6.2
```
jchodera: Maybe just change this to `version: 0.6.2`?

schema/meta.yaml
```diff
@@ -1,15 +1,11 @@
...
-  fn: schema-0.5.0.tar.gz
-  url: https://pypi.python.org/packages/4b/eb/fadd6bc4967fded31495ea70aebd3bdc90b83e3944d08bfd841039062f83/schema-0.5.0.tar.gz
-  md5: cee375ebfa54f1c4a218438b3b1c6803
-
-build:
-  skip:
-    - [py3k]
+  fn: schema-0.6.2.tar.gz
+  url: https://pypi.python.org/packages/9b/3c/5924163d693f156684e1502df0b408512e18c70d7fa80244496fafdb5988/schema-0.6.2.tar.gz
+  mdf: ac1b359c31e9b37fbe9333902f6bdf76
```
jchodera: This should be `md5`, right?

jchodera:
Hooray!

schema/meta.yaml
mpharrigan: It would need to be quoted `version: "0.6.2"`

mpharrigan:
In this case it should be fine, but in the future, version strings should be quoted so there's no risk of them being parsed as floats

mpharrigan:
```python
In [11]: yaml.load('version: 0.6.2')
Out[11]: {'version': '0.6.2'}

In [12]: yaml.load('version: 0.6')
Out[12]: {'version': 0.6}
```

PR #585, opened 2016-09-16
[schema] 0.6.2 with python 3.x support

jchodera:
Replaces #584

### DIFF
--- a/autoprotocol/meta.yaml
+++ b/autoprotocol/meta.yaml
@@ -10,7 +10,6 @@ source:
    # List any patch files here
    # - fix.patch
 
-
 # build:
   # noarch_python: True
   # preserve_egg_dir: True
--- a/autoprotocol/meta.yaml
+++ b/autoprotocol/meta.yaml
@@ -10,6 +10,7 @@ source:
    # List any patch files here
    # - fix.patch
 
+
 # build:
   # noarch_python: True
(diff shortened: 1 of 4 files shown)


jchodera:
For some reason, `conda-build-all` is failing with
```
This script requires conda-build and anaconda-client and
must run in the root conda environment.
  $ source deactivate
  $ conda install conda-build anaconda-client
```
It looks like `conda-build` version `1.x` may no longer be supported. I'm testing with the latest version right now.

jchodera:
Actually, that's probably not it. Having some more debug info printed.

Lnaden:
So I think the problem is with [This line](https://github.com/jchodera/conda-recipes/blob/b2df9c53f1e8feeb037648613ebb75aed421e0ee/conda-build-all#L30) importing the `hide_binstar_tokens` function. It was removed from the `conda/config.py` module in [this change in mid July](https://github.com/conda/conda/commit/7318479d11ece2f04c9d8b44bdba187868c593a9) in conda and moved to the `conda/base/context.py` module. 

The good news is, we don't seem to use that function anywhere in the `conda-build-all` script, so removing it seems to fix the the build as the #584 now passes with that line commented out. It may be better to use this PR instead so we don't have duplicate PRs.

jchodera:
Superceeded by #584

PR #586, opened 2016-09-16
Made version number string again in schema

Lnaden:
Fix leftover issues from #584.

@mpharrigan Patching this here should correctly replace the build on the omnia conda channel, yes?

### DIFF
--- a/schema/meta.yaml
+++ b/schema/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: schema
-  version: 0.6.2
+  version: "0.6.2"
 
 source:
   fn: schema-0.6.2.tar.gz


mpharrigan:
Since you had 
```yaml
version: 0.6.2
```
with two periods, it correctly interpreted it as a string. So the build is completely fine in this case! 

I want to make sure the convention is to always put it in quotes, just in case we ever have a build version like
```yaml
version: 0.6
```
This would be interpreted as a number (which is wrong). 


In this case, the package is completely fine. I'd be +1 merging this change so anyone looking at this recipe in the future will see the correct convention. It won't re-build the package because it already exists. If you want to have the package rebuilt, you have to increment the build number (which is just another field in the `.yaml` file), but a rebuild isn't required :)

Lnaden:
I figured that was the case. Thanks for clarifying. I'm also +1 on merging this if nothing else than for those looking at it in the future.

PR #591, opened 2016-09-21
Switch to conda build 2.0.1

marscher:
Changed conda-build-all to support conda-build 2.0 and pinned it in the scripts via the env variable "CONDA_BUILD", so we can easily change that.

Rationale: the new version supports to get version numbers dynamically and it provides lots of bug fixes. 

However we need to carefully test this. I ran several tests locally and it works so far. 

Should I also add a new dummy recipe in this PR (temporarily) to prove it does what it should?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ services:
 env:
   global:
     - CONDA_BUILD: "2.0.1"
-    secure: "GjbAskNQDToMpTQ87p2TcG+JPdr+TIyWxarkzwd/KCQLuJqTKVVinWtKZwB+uaEnz2mmnXT7jFgbUU2Vuq0Dv78SOgTnqeh+ObP60imN4z8TB88PAFihv7JDJ45Nfbe7viKrWLWDYRebkObbJduc7xms0YEb9vs4e1+knTojiEc="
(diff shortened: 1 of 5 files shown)


marscher:
I do not understand why skipping the recipe path in c924a8f does not continue with the other paths...

marscher:
Ah no it works as expected. I got confused by the Appveyor output, since it is not verbose. The Linux build failed since the Docker img could not be downloaded (maybe Dockerhub reduced the bandwidth of this image, since I pulled it too often)...

marscher:
https://travis-ci.org/omnia-md/conda-recipes/jobs/161637205 all the recipes are correctly being checked for availability on Anaconda cloud.

jchodera:
Awesome progress!

I just restarted the travis build.

marscher:
good luck with that.. DockerHub is deadly slow at moment.

jchodera:
Maybe we should have a "lightweight" build image as well as a "heavyweight" image (including OpenCL + CUDA + TeX).

Some of the Pande group folks are already testing conda-forge builds, which is another alternative for packages like `pyemma`. We could probably migrate many of our projects there, and assist them in getting a framework up for building things like `OpenMM` to migrate the rest.

marscher:
Am 21.09.2016 um 18:09 schrieb John Chodera:
> Maybe we should have a "lightweight" build image as well as a "heavyweight" image (including OpenCL + CUDA + TeX).
>
> Some of the Pande group folks are already testing conda-forge builds, which is another alternative for packages like |pyemma|. We could probably migrate many of
> our projects there, and assist them in getting a framework up for building things like |OpenMM| to migrate the rest.
Yes, its probably much less of a hassle to us, if we can get rid of all this technical maintenance.

jchodera:
Everything passes!

jchodera:
Thanks again!

PR #592, opened 2016-09-21
[pyemma] bump 2.2.5

marscher:


jchodera:
Should we merge #591 first?

marscher:
nevermind

PR #596, opened 2016-09-24
[yank] 0.10.0 (with python3 support)

jchodera:


jchodera:
We should pin the `alchemy` version to 1.3.0 after its release. 

Don't merge this yet.

jchodera:
Note that we can't install `mpi4py` by default on `win` systems since travis tests won't pass:
```
--------------------------------------------------------------
mpi4py requires that Microsoft MPI 6 be installed on the host
system. That is a systemwide installation that is currently
not available through conda. In order to successfully use
mpi4py you must install Microsoft MPI and then append the bin
directory of the MPI installation to your PATH environment
variable. To install Microsoft MPI see
https://www.microsoft.com/en-us/download/details.aspx?id=47259
--------------------------------------------------------------
```

jchodera:
Looks like we can't build on `win` yet, so we'll skip it for now.

PR #604, opened 2016-10-04
Updating YANK to version 0.11.2

Lnaden:
Updates YANK to 0.11.0

- Version number is now a string
- Fixed Schema dependency


Lnaden:
Not passing due to missing `tlmgr`. Was there a stealth update to TexLive causing something to break here?

PR #613, opened 2016-10-16
[parmed] 2.6.1

jchodera:


jchodera:
cc: https://github.com/ParmEd/ParmEd/issues/768

PR #616, opened 2016-10-21
[pyemma] bump 2.2.7

marscher:


jchodera:
Sorry for the delay in merging this. The DDoS attack yesterday rendered GitHub essentially unusable from the east coast.

Everything looks good!

PR #619, opened 2016-10-24
bump msmtools 1.2

marscher:


jchodera:
Let me quickly fix the sams issue I accidentally introduced into appveyor before merging this

PR #622, opened 2016-10-28
removed pytest

marscher:
3.0.3 is available upstream and there is no need to duplicate it here.


jchodera:
Looks like the failure was the `PyQT4` issue again. I retriggered the build, but we can merge after that even if the `PyQT4` failure occurs again since this seems to be unrelated.

PR #623, opened 2016-10-28
Update to the AmberMini 16.16.0, based on AmberTools 16 patch 16

Lnaden:


jchodera:
I think you have to add
```YAML
  skip:
    - [win]
```
to the `build:` section.

PR #625, opened 2016-11-01
Critical fix for the  MD5 issue for ambermini

Lnaden:
The MD5 hash on the github release tarbal keeps changing, this is causing all builds of ambermini to fail. This removes the MD5 check and updates build number to hopefully get the omnia channel to update is release

### DIFF
--- a/ambermini/meta.yaml
+++ b/ambermini/meta.yaml
@@ -5,10 +5,9 @@ package:
 source:
   fn: 16.16.0.zip
   url: https://github.com/choderalab/ambermini/archive/16.16.0.zip
-  md5: 2969b5b042a0ebd83ead860a9a8c4d2e
 
 build:
-  number: 5
+  number: 6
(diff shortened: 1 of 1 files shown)


Lnaden:
The ambermini build is okay, msmexplorer is still causing issue on the py34 and py35, but this should fix the ambermini build at least

mpharrigan:
Shouldn't we be concerned that the md5 hash is changing?

Lnaden:
Yes, that is a concern I have, but I cannot reliably recreate the issue and I also can't find evidence of someone else having the same problem. All I know is the MD5 of the tarball right after I cut the release changed about 3 times at different download points, but the last two samples (yesterday and today) are the same. I have not reached out to github yet to see if this is something they have encountered before, but for the purposes of getting a stable AmberMini v16 out, I figured removing the MD5 was the fastest for now.

PR #626, opened 2016-11-01
remove msmexplorer tests

cxhernandez:
re: https://github.com/msmexplorer/msmexplorer/issues/68


cxhernandez:
Not sure why the Travis build isn't being triggered

cxhernandez:
@jchodera: removing the import tests works. not sure how you feel about this

jchodera:
Huh. Sure, this sounds like a reasonable workaround for now. Thanks!

PR #627, opened 2016-11-02
[msmexplorer] Disable builds, since they were causing omnia PRs to fail

jchodera:


jchodera:
Closed in favor of #626

PR #636, opened 2016-11-11
Debug osx basictex issues

jchodera:


jchodera:
Might be [this issue](https://github.com/caskroom/homebrew-cask/issues/6179)

PR #638, opened 2016-11-11
MDTraj on conda-forge

mpharrigan:


jchodera:
Thanks for piloting this!

jchodera:
The `osx` failure is unrelated, and no recipes are built by this PR, so merging.

Issue #634, opened 2016-11-08
MDTraj on conda-forge


mpharrigan:
Conda forge has some good infrastructure set up for community-maintained conda packages. https://conda-forge.github.io/

@rmcgibbo already has mdtraj set up to be built there. For our forthcoming release of 1.8, I plan on using the conda-forge machinery to build our official binaries and copy them over to the omnia channel. The omnia machinery won't be used to build the binaries

mpharrigan:
crossref https://github.com/mdtraj/mdtraj/issues/1196

jchodera:
I'm very interested in trying to migrate omnia over to conda-forge, and I think there is a path to do this with openmm as well. 

Do you foresee any issues with compatibility if you build on conda-forge? Which range of glibc libraries will the compiled code be compatible with, for example, and will that match the rest of omnia (which is compiled with CentOS 5.x)?

rmcgibbo:
The forward compatibility is fine, but not the backward compatibility, so these binaries will not be usable on CentOS 5.

rmcgibbo:
The only circimstance I can think of in which this might matter (other than for users who are still using EOL-ed linuxes) is that if one of the omnia project's build systems needed to use MDTraj's C/C++ code _at build time_ to compute some distances or RMSDs, it would break.

I can't think of any projects whose setup.py script requires running RMSD calculations though.

jchodera:
Sounds reasonable.

We should discuss potential migration to conda-forge in the next omnia call.

I don't see the harm in copying over the package to the omnia channel, but
I haven't thought through all the potential ramifications yet.

mpharrigan:
For those keeping score at home, this turned out to be easy.

https://docs.continuum.io/anaconda-cloud/cli#copy

```bash
$ anaconda login
$ anaconda copy conda-forge/mdtraj/1.8.0 --to-owner omnia
Using Anaconda API: https://api.anaconda.org
Copied file: linux-64/mdtraj-1.8.0-np110py27_0.tar.bz2
Copied file: linux-64/mdtraj-1.8.0-np110py34_0.tar.bz2
Copied file: linux-64/mdtraj-1.8.0-np110py35_0.tar.bz2
Copied file: linux-64/mdtraj-1.8.0-np111py27_0.tar.bz2
Copied file: linux-64/mdtraj-1.8.0-np111py34_0.tar.bz2
Copied file: linux-64/mdtraj-1.8.0-np111py35_0.tar.bz2
[...]

PR #645, opened 2016-12-01
[openpathsampling] 0.9.2

dwhswenson:
[OpenPathSampling 0.9.2](https://github.com/openpathsampling/openpathsampling/releases/tag/v0.9.2)

### DIFF
--- a/openpathsampling/meta.yaml
+++ b/openpathsampling/meta.yaml
@@ -36,9 +36,9 @@ requirements:
     - matplotlib
     - ujson
 
-test:
-  imports:
-    - openpathsampling
+#test:
+  #imports:
+    #- openpathsampling
 
 about:
   home: http://github.com/openpathsampling/openpathsampling
--- a/openpathsampling/meta.yaml
+++ b/openpathsampling/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: openpathsampling
-  version: 0.9.1
+  version: 0.9.2
 
 source:
   git_url: https://github.com/openpathsampling/openpathsampling.git
-  git_tag: v0.9.1
+  git_tag: v0.9.2
 
 build:
   preserve_egg_dir: True
@@ -27,11 +27,12 @@ requirements:
     - netcdf4
     - openmm
     - openmmtools
+    - jupyter
     - pyyaml
     - mdtraj
     - svgwrite
     - networkx
-    - pyqt 4*
+    - pyqt 4.11.4 py27_3
     - matplotlib
     - ujson
 


dwhswenson:
I notice that the appveyor error is the same as in #644 and #643. The Travis error is the same as #644, but #643 passed fine. Any suggestions on how to fix? The only changes are in our `meta.yaml`: a new req, a specific pin, and a version bump. Our tests build on Travis with a local version of our package, and otherwise the same `meta.yaml`.

jhprinz:
Seems to be a problem when importing mdtraj. Not much we can do about it. It says `glibxx` missing or so...

jhprinz:
```
ImportError: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.11' not found (required by /anaconda/conda-bld/openpathsampling_1480633113236/_t_env/lib/python2.7/site-packages/mdtraj/formats/dtr.so)
```

Is it possible that the fixing of `qt` causes a problem with the latest mdtraj?

rmcgibbo:
This is exactly https://github.com/omnia-md/conda-recipes/issues/634#issuecomment-259250071. cc @mpharrigan

jchodera:
Thanks, @rmcgibbo!

@dwhswenson and @jhprinz, do you need to run mdtraj at build time? If so, it looks like we have to revert to building mdtraj within omnia for the omnia version.

jhprinz:
No, we don't need to build it. Not sure why it tries to, we just need to conda install it and then it is used in a test.

dwhswenson:
Does the script here test that our code can be correctly imported? (`test: imports: openpathsampling` in the meta.yaml). MDTraj is imported when you run `import openpathsampling` (under a layer or two of other imports, but certainly inside of `openpathsampling.engines.openmm.tools` and `openpathsampling.collectivevariable`). Is that the problem?

jchodera:
If you disable the `import` test, does it succeed? I don't necessarily like doing that, but it would get this built for now.

mpharrigan:
Sorry about this. As noted above, the mdtraj conda packages are built on conda-forge which is based off of rhel/centos 6, whereas omnia uses rhel/centos 5. We could continue building two sets of binaries. If the only time this crops up is in a docker container we explicitly picked to be super old, I'd rather just say "mdtraj is centos >= 6"

jchodera:
> We could continue building two sets of binaries

I actually think this is the best approach for now until we fully switch the build system to something that is based on CentOS 6 in preparation for migrating to conda-forge.

>  If the only time this crops up is in a docker container we explicitly picked to be super old, I'd rather just say "mdtraj is centos >= 6"

As illustrated here, that either means (1) we can't test that omnia tools that depend on `mdtraj` correctly built (which worries me a bit), or (2) you are excluding any `omnia`-built tools from using `mdtraj`. I think it may be OK to go with (1) for the very short term, though I worry this will cause trouble with quality control.

mpharrigan:
Since the quality control is just importing the package, we could probably just do that outside the docker container.

For certain definitions of "short term" (3 months), continuing to build packages on centos 5 is a short term solution. Its EOL is March 31, 2017. We shouldn't be distributing binaries built on an unsupported OS. Does holy build box have a migration planned? Maybe this is an opportunity to start moving things to conda-forge? It would probably be worth-while to ask if you can submit a conda-forge recipe that uses the omnia openmm package (since figuring out how to build that on a new platform will be a headache).

mpharrigan:
A very timely PR: https://github.com/phusion/holy-build-box/pull/3

jchodera:
> For certain definitions of "short term" (3 months), continuing to build packages on centos 5 is a short term solution. Its EOL is March 31, 2017. We shouldn't be distributing binaries built on an unsupported OS.

I totally agree.

> Does holy build box have a migration planned? 

Looks like we can upgrade to the new CentOS 6 version soon, as you note.

> Maybe this is an opportunity to start moving things to conda-forge? 

I think it's great to do this, but it will be made *much* easier when `omnia` and `conda-forge` are both on CentOS6 and packages are cross-compatible during the migration.

> It would probably be worth-while to ask if you can submit a conda-forge recipe that uses the omnia openmm package (since figuring out how to build that on a new platform will be a headache).

Definitely, now that we're not just getting a `nan` sandwich with CUDA 8!

dwhswenson:
> If you disable the `import` test, does it succeed? I don't necessarily like doing that, but it would get this built for now.

It passed Travis (Linux) when I removed the `import` test from the `meta.yaml` (waiting on OS X; [apparently it is having severe delays today](https://www.traviscistatus.com/incidents/nq0sf0srvx8s)).

Still fails AppVeyor, but I think that's a more general problem. (OPS is supposed to be skipped in the Windows tests, plus I see the same error in other PRs and in the most recent nightly.)

> Since the quality control is just importing the package, we could probably just do that outside the docker container.

Is this something I should do in this PR, or does it reflect a larger structural change in the whole of `omnia/conda-recipes`? (in which case I assume it is someone else's task)

jchodera:
> Is this something I should do in this PR, or does it reflect a larger structural change in the whole of omnia/conda-recipes? (in which case I assume it is someone else's task)

This is a larger structural issue that needs to be addressed.

I'll look into the AppVeyor failures separately.

PR #651, opened 2016-12-08
Re-format skip sections

mpharrigan:
Per http://conda.pydata.org/docs/building/meta-yaml.html#skipping-builds I've put all the skips as comments.

I've put them on one line, so you can grep for "skip: " and get a quick summary of what's not getting built.

If there's a comment explaining the skip, put it on one line before (so you can `grep -B1`).

We should make people explain their skips. There are a lot of skipped configurations whose reasoning is not obvious at all. This is important if someone wants to fix it


jchodera:
Or should this be merged first?

mpharrigan:
Let's do #650 first, then this, then leave #652 open until everything supports py3

PR #652, opened 2016-12-08
Enable py3

mpharrigan:
per #551 , remove python 3 skips (which will cause travis to fail). As people fix python 3 compatibility, things will fail less


jchodera:
Didn't get to this quickly enough due to NIH visit today. Can you fix the conflicts and then we can merge?

Sorry about merging #653 before this!

mpharrigan:
Let's not actually merge this (since it will cause travis to fail). I'll keep rebasing as people make progress towards python 3 compatibility

jchodera:
@mpharrigan : Can you rebase/merge from `master` when you get a chance? We can get all of this cleaned up pretty quickly thereafter.

mpharrigan:
rebased. This shouldn't be merged. The packages noted here should work to support py3

jchodera:
Thanks!

PR #657, opened 2016-12-23
Update build infrastructure to CUDA 8.0; openmm 7.1.0 beta

jchodera:


marscher:
@jchodera When you removed the noarch flag, you should also have remove the build skip section...

jchodera:
> @jchodera When you removed the noarch flag, you should also have remove the build skip section...

Good catch! Sorry about that. Will fix.

PR #661, opened 2017-01-01
[openmmtools] 0.8.3

jchodera:


conda-build-all
```diff
@@ -256,8 +256,7 @@ def upload_to_binstar(m, python, numpy, username, max_retries=5,
     cmd.extend(['upload', '--no-progress', '-u', username, filename])
     if force:
         cmd.extend(['--force'])
-    # Force upload if extra: force_upload: True is specfied
-    elif (m.get_section('extra') and ('force_upload' in m.get_section('extra')) and (m.get_section('extra')['force_upload'])):
+    elif m.get_section('extra') and ('force_upload' in m.get_section('extra')) and (m.get_section('extra')['force_upload']=='True'):
```
rmcgibbo: are you sure that force_upload is of type `str` with value 'True`'? I thought it was a bool.

jchodera: Much to my surprise, this expression evaluated to `True` despite the YAML file containing `force_upload: False`. I'll double check its type and fix in a separate PR.

rmcgibbo: Yaml and python are slightly different in how your write booleans. It should be `false`, not `False` in the YAML file.


http://www.yaml.org/refcard.html: ```
Language Independent Scalar types:
    { ~, null }              : Null (no value).
    [ 1234, 0x4D2, 02333 ]   : [ Decimal int, Hexadecimal int, Octal int ]
    [ 1_230.15, 12.3015e+02 ]: [ Fixed float, Exponential float ]
    [ .inf, -.Inf, .NAN ]    : [ Infinity (float), Negative, Not a number ]
    { Y, true, Yes, ON  }    : Boolean true
    { n, FALSE, No, off }    : Boolean false
```

jchodera: Wait, that reference card doesn't say `false` is an option. Only one of `{ n, FALSE, No, off }`. Am I missing something?

jchodera: Or are both `false` and `FALSE` permitted, but not `False`?

jchodera: [This specification](http://yaml.org/type/bool.html) suggests that `False` is permitted.

jchodera: I'm not sure why, but empirically, it seems that both `false` and `False` become `str` type---at least locally, on `osx` Python 3.5.

rmcgibbo: Perhaps it's something else that's causing the problem then.

PR #663, opened 2017-01-06
Enable python 3.6

mpharrigan:


mpharrigan:
It's just trying to build the first package on a loop

marscher:
Is Python 3.4 also deprecated?

jchodera:
Let's try to follow what conda-forge is doing here.

marscher:
I honestly dont think this software channel will survive long enough to clarify the effort of supporting this brand new version here. Conda-forge uses a RC of 3.6 so far and the upgrade process has not yet triggered all feedstock to build for it, so it is still far away from being usable.

mpharrigan:
> Is Python 3.4 also deprecated?

Per our discussion when 3.5 was released, we'll build new packages for 3.5 and 3.6 (latest two releases). The existing python 3.4 binaries won't be going anywhere

rmcgibbo:
From our readme:

```
Supported versions

Python packages are built against latest two releases of python and python 2.7. Packages which have a binary dependency on Numpy are built against the latest two releases of Numpy.
```

jchodera:
Looks like anaconda now has python 3.6 support:
https://www.reddit.com/r/Python/comments/5ox7ot/python_36_now_in_anaconda_repo/

Does anyone know what conda-forge is currently building with?

jchodera:
We're now building for python 3.6 via #698!

PR #664, opened 2017-01-08
[progress_reporter] Remove skip for different platforms since we no longer use noarch

jchodera:
Fixes issue noted by https://github.com/omnia-md/conda-recipes/pull/657#issuecomment-270964966


jchodera:
We still need to sort out that appveyor error, but this seems to be working as intended now.

PR #668, opened 2017-01-14
Remove funcsigs recipe

jlmaccal:
Previously, we patched `1.0.0` of `funcsigs` because it did not
correctly handle the dependency on `ordereddict` in newer versions
of python. Upstream now handles this correctly, so this patch
should no longer be required.


jchodera:
Thanks! Looks like it's OK to leave the existing `funcsigs` package for reproducibility because new `funcsigs` installs will grab 1.0.2 from the [anaconda channel](https://anaconda.org/anaconda/funcsigs/files).

I just noticed that some of my nightly builds were failing because of failure to import `funcsigs` from `mock`:
https://travis-ci.org/choderalab/ensembler/jobs/191780038#L1168-L1184

I'm not sure why this is suddenly happening, though.

jlmaccal:
The version from `omnia-md/conda-recipes` is actually not building / installing correctly, which is why your tests that use `mock` are failing. I was having the same problem with MELD.

I'm not sure what you mean about new installs grabbing 1.0.2. The `omnia-md` channel overrides anaconda, so we're getting the patched version 1.0.0, which a I mentioned above, doesn't build correctly.

jlmaccal:
> I'm not sure why this is suddenly happening, though.

I suspect this is due to anaconda recently (last few days) updating to a new version of conda and related packages.

jchodera:
> The version from omnia-md/conda-recipes is actually not building / installing correctly, which is why your tests that use mock are failing. I was having the same problem with MELD.

This means we need to both remove the recipe and delete the `funcsigs` package from the `omnia` channel, because it is broken?

> I'm not sure what you mean about new installs grabbing 1.0.2. The omnia-md channel overrides anaconda, so we're getting the patched version 1.0.0, which a I mentioned above, doesn't build correctly.

My understanding was that the latest version from any available channel would be pulled, but maybe that's incorrect.

jlmaccal:
> This means we need to both remove the recipe and delete the funcsigs package from the omnia channel, because it is broken?

Yes, that's what I am suggesting.

> My understanding was that the latest version from any available channel would be pulled, but maybe that's incorrect.

I'm not sure how it resolves them, but it's definitely pulling from the `omnia` channel:
```
    ensembler-dev:   0.0.0-py27_0       local          
    fastcluster:     1.1.20-np110py27_0 omnia          
    fftw3f:          3.3.4-2            omnia          
    freetype:        2.5.5-1                           
    funcsigs:        1.0.0-py_1         omnia          
    hdf5:            1.8.16-0                          
    hdf5-1814:       1.8.14-0           omnia          
    imagesize:       0.7.1-py27_0                      
    jbig:            2.1-0                             
    jinja2:          2.8.1-py27_0                      
    jpeg:            9b-0                              
    libgfortran:     1.0-0                             
    libiconv:        1.14-0    
```

jlmaccal:
I've poked around at this for too long. Conda is doing something weird, although I haven't yet figured out why.

In any case, I think we should remove this. It only exists to support `python<=2.5`, which we don't support anyway.

@jchodera Can you merge this in and remove `funcsigs` from the omnia channel. Once that is done, I'll figure out the other issue with the latest conda update.

jchodera:
I've moved the `funcsigs` package to the `deprecated` label, just in case we need it in the future for reproducibility:

https://anaconda.org/omnia/funcsigs/files

Let me know if it still gets picked up by your installs and I can remove it.

PR #670, opened 2017-01-14
[WIP] Fix conda build

jlmaccal:


jlmaccal:
@jchodera This appears to be working, but Travis kills the build because the log file is too large.

jchodera:
Thanks! Looks like the `osx` build is still having some problems, though.

I'll see if I can start with these changes in a separate PR.

jlmaccal:
> Thanks! Looks like the osx build is still having some problems, though.
> 
> 
Yeah, I hadn’t even tried to make any changes to anything other than linux yet.

jlmaccal:
I can confirm that it is now pulling the correct version.

jlmaccal:
@jchodera : This is now correctly skipping all of the up-to-date builds. I just had to extract the filename and channel from a structure, rather than just splitting a string.

I also temporarily bumped the version of one of the packages (eigen3) to ensure that changed packages were indeed being built. If you want to incorporate the changes into your version, the relevant change is on 314/315 of `conda-build-all`.

jchodera:
Thanks again for solving this! We merged #671 instead in order to fix `osx` builds as well.

Lnaden:
The `osx` travis-ci builds have not run yet, so we only have the linux compiled versions on the `omnia` conda right now. There is currently an `osx` backlog on travis. The builds should run as the backlog clears out, so we may still see issues as people retry their own software builds.

Lnaden:
Now they are up there. We should be good to go on re-running other projects

PR #671, opened 2017-01-15
Don't pin conda-build to 2.0.1 due to API incompatibilities with 'conda'

jchodera:
Supercedes #670 by updating `linux`, `osx`, and `win` builds (though there are likely other problems with the appveyor build).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ services:
 
 env:
   global:
-    - CONDA_BUILD: "2.0.1"
     - CUDA_VERSION: "8.0"
     - CUDA_SHORT_VERSION: "80"
   secure:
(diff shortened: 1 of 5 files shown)


jchodera:
@jlmaccal : Your fix to `conda-build-all` makes it want to rebuild *all* packages, which indicates to me that something is very wrong. We certainly don't want that to happen!

jlmaccal:
Ugh... I wish I knew what `get_bldpkg_path` returned previously...

The issue is in the logic in `required_builds`, where `base_buildpkg_path` contains the suffix `.tar.bz2`, while the keys in `index_by_pkgname` do not.

jlmaccal:
So, from looking at old Travis logs, it seems that `get_bldpkg_path` returns exactly the same paths as before.

I think the issue is with other changes in conda, where the key to dictionaries returned from `conda.api.get_index` used to be a string, but now has type `conda.models.dist.Dist`. So the comparisons in `required_builds` are guaranteed to fail.

jchodera:
What if we just pin Miniconda to a slightly earlier version that is compatible with conda-build 2.0.1 until we can figure this out?

jlmaccal:
Freezing the version of `conda` and `conda-build` is certainly a viable short term option, but presumably we’ll need to figure this out eventually.

Lnaden:
This looks like its passing barring the AppVeyor builds which are all queued and not running. Do we want to merge this version so we can get new builds out?

On a related note, are we going to have to increase the build number of all packages to get them to rebuild on the newer `conda-build` and push out to `omnia`'s `conda` channel?

jchodera:
Go ahead and merge if it looks good!

> On a related note, are we going to have to increase the build number of all packages to get them to rebuild on the newer conda-build and push out to omnia's conda channel?

We'll only do that for packages which are not working as a result, like `ambermini`

Issue #675, opened 2017-01-18
Need to rebuild swig


jlmaccal:
I get the following error when trying to build MELD:

```
conda.CondaMultiError: Placeholder of length '80' too short in package /anaconda/conda-bld/master_1484775469603/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/bin/swig.
The package must be rebuilt with conda-build > 2.0.
```

Swig is pulled in from the omnia channel.

@jchodera @Lnaden: Can we bump the build number of swig to force a rebuild?

Incidentally, why are we packaging our own version of swig anyway? Isn't included as part of anaconda?

jchodera:
@Lnaden: Can you bump the `swig` *build number*?

@jlmaccal: I'm not sure `swig` was always available---`swig` 3.0.8 was only uploaded 11 months ago, and 3.0.7 (which was used to build OpenMM  7.0.1) isn't available at all via `conda`.

Lnaden:
Yes. 

I am trying to actually replicate this issue locally so we can proactively find these instead of reactively.

jlmaccal:
OK...

I was building against 3.0.7,  which was likely built with conda < 2.0.

I've unpinned the version of swig used in MELD, which should pull in the latest version of swig, presumably built with conda 2.0.1 or later.

It occurs to me that one weakness of the current system is that changes to the build tooling (meta.yaml, docker_build.sh, et al) don't actually trigger a rebuild of the packages.

jchodera:
> I am trying to actually replicate this issue locally so we can proactively find these instead of reactively.

I think the cascade of `nightli.es` will clue us in to any other issues like this over the next couple of days.

jchodera:
> It occurs to me that one weakness of the current system is that changes to the build tooling (meta.yaml, docker_build.sh, et al) don't actually trigger a rebuild of the packages.

They *shouldn't*. And they shouldn't need to. It's total bullshit for continuum to break backward-compatibility like this.

jchodera:
If we really need to, we could use the `--force` flag once to force a rebuild of all packages, but it may not make it through all of them, and they would all have the same build number.

Let's try to be conservative and only rebuild the ones that need it by incrementing the build numbers.

Lnaden:
I made a PR for this by unpinning meld. 

[...]

PR #676, opened 2017-01-18
Unpin meld's swig version to unbreak from conda update

Lnaden:
Partially addresses #675

Still need to find a proactive way to catch these

### DIFF
--- a/meld-plugin/meta.yaml
+++ b/meld-plugin/meta.yaml
@@ -15,7 +15,7 @@ requirements:
     - openmm-dev
     - cmake
     - python
-    - swig ==3.0.7
+    - swig
     - eigen3
 
   run:


jlmaccal:
Sorry for the confusion...

The error was not due to the recent update of conda in the last few days. It was due to a previous update to conda 2.0, which happened in September.

The swig version for the meld openmm plugin was pinned at 3.0.7. This version of swig was built with an old version of conda-build. I have since unpinned the version of swig, so it now grabs 3.0.8, which was built with conda-build 2.0.1 and works fine.

jlmaccal:
@Lnaden No, not yet. Sorry for the confusion.

We actually build the release version of MELD here, but we have automated builds for testing every time there is a commit over in /maccallumlab/meld. That's what I've been working on. Once I have that fixed up, I'll update the build here.

You can probably close this PR.

Lnaden:
Not a problem, thanks for clarifying!

PR #677, opened 2017-01-20
Rebuild packages from conda-build break

Lnaden:
The following package was still broken from the conda-build kerfluffle. 
* deepchem

The following package could not build due to other `conda` related problems:
* parmed 
    * MD5 Mismatch, the tarball on GitHub changed its MD5 signature. 
    * If a new version is cut, the MD5 will be updated and wont be an issue. 
    * If we have to _rebuild_ the current version this will be a problem.
    * cc @swail package maintainer in case this is important. 

All other packages were able to `conda build` without the placeholder length error. There were a few that could not build due to various other errors that seem to be program specific, and not related to the `conda` problems.

Its worth noting that the `nightli.es` would never have caught this since it does not actually try rebuilding packages unless the version or build number is incremented.

### DIFF
--- a/deepchem/meta.yaml
+++ b/deepchem/meta.yaml
@@ -7,7 +7,7 @@ source:
     git_tag: 0.0.4
 
 build:
-  number: 0
+  number: 1
   skip: True  # [osx or win or py3k]
 
 
--- a/deepchem/meta.yaml
+++ b/deepchem/meta.yaml
@@ -3,7 +3,7 @@ package:
   version: "0.0.4"
 
 source:
-    git_url: https://github.com/pandegroup/deepchem/
+    git_url: https://github.com/pandegroup/deepchem.git
     git_tag: 0.0.4
 
 build:
--- a/openbabel/meta.yaml
+++ b/openbabel/meta.yaml
@@ -12,7 +12,7 @@ source:
     - fix_data_path.diff
 
 build:
-  number: 1
+  number: 2
   detect_binary_files_with_prefix: true
   skip: True  # [win]
 


Lnaden:
Apparently I am still missing some packages. I'll keep updating them as I find them

jchodera:
Tagging @rbharath re `deepchem`.

Can we just copy the [openbabel package](https://anaconda.org/openbabel/openbabel) to the omnia channel?

mpharrigan:
> Can we just copy the openbabel package to the omnia channel?

Do they support centos 5 :)

jchodera:
> Do they support centos 5 :)

Oof, good point.

rbharath:
Is DeepChem now building without problems? The version on here is a ways behind HEAD, so I'll probably be making a version bump soon so we update the package on here as well.

For DeepChem purposes, should be fine to just use Anaconda openbabel.

jchodera:
> Is DeepChem now building without problems? 

Nope. See https://github.com/omnia-md/conda-recipes/pull/679

Lnaden:
Once the OpenBabel package has been rebuilt and pushed to the `omnia` channel, I'll restart the build for this one.

Lnaden:
Okay. Slightly different problem. See #681

Lnaden:
Odd. The linux builds seem to have issues fetching the tarbal from github, but the osx ones appear to be working correctly, I've restarted the job once already and am running it again just in case it was something odd with GithHub. Has anyone seen this before?

Lnaden:
The unix builds work now, but the OSX builds do not because the 1.4GB CUDA files are taking too long to download from nvidia right now, causing travis to time out.

jchodera:
Go ahead and merge. I might be able to stage those files on S3

PR #679, opened 2017-01-20
Building openbabel first, will come back for deepchem

Lnaden:
#677 is not building because OpenBabel is not building. I think this is a chicken-egg problem so I am going to try and build and push openbabel first, then come back for deepchem.

### DIFF
--- a/openbabel/meta.yaml
+++ b/openbabel/meta.yaml
@@ -34,11 +34,16 @@ test:
   imports:
     - openbabel
     - pybel
+  source_files:
+    - test/test_pybel.py
+    - test/testkekule.py
+    - test/testbindings.py
+    - test/testexample.py
   commands:
-    - cd $SRC_DIR/test ; python test_pybel.py
-    - cd $SRC_DIR/test ; python testkekule.py
-    - cd $SRC_DIR/test ; python testbindings.py
-    - cd $SRC_DIR/test ; python testexample.py
+    - python test_pybel.py
+    - python testkekule.py
+    - python testbindings.py
+    - python testexample.py
 
 about:
     home: http://www.openbabel.org/
--- a/openbabel/meta.yaml
+++ b/openbabel/meta.yaml
@@ -12,7 +12,7 @@ source:
     - fix_data_path.diff
 
 build:
-  number: 1
+  number: 2
   detect_binary_files_with_prefix: true
   skip: True  # [win]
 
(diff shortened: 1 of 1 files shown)


jchodera:
@Lnaden : It looks like everything works in building `openbabel` except the test script path is misconfigured (or the test file is not checked in):
```
===== testing package: openbabel-2.4.0-py35_2 =====
import: 'openbabel'
import: 'pybel'
running run_test.py
===== openbabel-2.4.0-py35_2 OK =====
+ /bin/bash -x -e /anaconda/conda-bld/openbabel_1484940011820/test_tmp/run_test.sh
+ cd /anaconda/conda-bld/openbabel_1484940011820/test_tmp/test
/anaconda/conda-bld/openbabel_1484940011820/test_tmp/run_test.sh: line 3: cd: /anaconda/conda-bld/openbabel_1484940011820/test_tmp/test: No such file or directory
TESTS FAILED: openbabel-2.4.0-py35_2
```

Lnaden:
I'm not exactly sure how to fix that. I've got a few things I can try. Altenetley, we can remove the `test` section from the `meta.yaml` file and rely instead on the `run_test.py` file for testing. 

From what I can tell, it looks like something changed in the way the source is built on the back end. Its been a year since this was originally compiled, so I imagine some things changed under the hood which are now causing this to fail.

jchodera:
Let's just remove [these lines](https://github.com/Lnaden/conda-recipes/blob/6f1c26a7bf73427f48675a216ba3cbcc01a64444/openbabel/meta.yaml#L37-L46).

Lnaden:
Sure, I just changed those from what they were to try and work around the error. If we remove those we reduce the ability to test the package though.

jchodera:
> Sure, I just changed those from what they were to try and work around the error. If we remove those we reduce the ability to test the package though.

@rbharath can improve the testing in a separate PR if he ends up finding needed functionality is missing.

Lnaden:
There is a test that is still run which does [do the imports](https://github.com/Lnaden/conda-recipes/blob/df9189979290c85217df417368a01032054b1a63/openbabel/run_test.py#L8) (PyBel imports the OpenBabel libs). So I think having the `meta.yaml`'s `test` section do an import of the two packages is redundant, if I am interpreting the `run_test.py` script correctly. 

I can also verify from the travis builds that the `run_test.py` script [is being run successfully](https://travis-ci.org/omnia-md/conda-recipes/jobs/194458063#L4354-L4357), so I think we are good.

jchodera:
Ah, OK. As long as there is some minimal check of whether it built the package, that should be fine!

PR #683, opened 2017-01-24
[openmm] 7.1rc1

jchodera:
cc: https://github.com/pandegroup/openmm/issues/1718


jchodera:
I don't understand how or why, but these ended up being pushed to the `dev` channel instead of `rc`, so I've moved the manually.

peastman:
And the Windows one ended up in the `main` channel!  I've fixed it.

mpharrigan:
> I don't understand how or why, but these ended up being pushed to the dev channel instead of rc, so I've moved the manually.

Probably https://github.com/omnia-md/conda-recipes/blob/master/conda-build-all#L260

jchodera:
Ah! `upload`, if specified, should take precedence. We just have to reorder those if branches.

PR #686, opened 2017-01-26
[openmm] Try building osx sphinx manual with make -j1

jchodera:


jchodera:
OK, that approach obviously doesn't work.

PR #687, opened 2017-01-26
Fix doxygen segfault on osx

jlmaccal:
Revert to previous version of doxygen from homebrew.


jchodera:
Ooh, good idea!

jchodera:
It works! Thanks!

PR #689, opened 2017-02-01
Update Yank to 0.15.0

Lnaden:


jchodera:
Should be OK to merge.

mpharrigan:
I thought we typically wait for the tests to pass before merging? Why isn't appveyor running? Did you notice the osx import test fails?

Lnaden:
The AppVeyor builds have not been running for almost 3 weeks now (they just sit in Pending forever), and they haven't been passing for 3 months when they were running. I'm not sure what is going on there as there does not appear to be any bottleneck I can see, but I also have not had any time to really pin down what is going on there. 

I can revert these changes if you want.

jchodera:
> I thought we typically wait for the tests to pass before merging? Why isn't appveyor running? Did you notice the osx import test fails?

My bad. I totally didn't realize this was the conda-recipes repo. Sorry about that---github notification overload today.

> The AppVeyor builds have not been running for almost 3 weeks now (they just sit in Pending forever), and they haven't been passing for 3 months when they were running. I'm not sure what is going on there as there does not appear to be any bottleneck I can see, but I also have not had any time to really pin down what is going on there.

Maybe we should pay for AppVeyor for a little while? It's only $29/mo for the basic plan.

Issue #690, opened 2017-02-01
Travis CI builds of master always report success


mpharrigan:
Even when packages fail to build, it reports success because the script is
```bash
conda-build-all ... || true
```

Why are we doing this?

jchodera:
We definitely want to do that on `master` since otherwise a broken package will mean that the build will *stop* and not build any subsequent packages.

However, on PRs, we don't want to do that---we want the build to fail if there are broken packages.

mpharrigan:
> the build will stop and not build any subsequent packages.

I don't think that's quite right. The `|| true` isn't inside a loop. It just modifies the exit code of conda-build-all. If conda-build-all exits early (on first failed package) the `|| true` isn't going to help

https://github.com/omnia-md/conda-recipes/blob/master/devtools/docker-build.sh#L17

The right place to implement this logic is in the package building loop *inside* conda-build-all.

mpharrigan:
Note that this logic *is* already implemented in conda-build-all. The exit code is set to the number of failed packages

jchodera:
> The right place to implement this logic is in the package building loop inside conda-build-all.

Ah! Agreed!

jchodera:
> Note that this logic is already implemented in conda-build-all. The exit code is set to the number of failed packages

OK, let's remove this now.

jchodera:
The `nightli.es` appear to have revealed a new issue now:
```
Fetching package metadata ....
WARNING: The remote server could not find the noarch directory for the
requested channel with url: file:///anaconda/conda-bld/noarch
It is possible you have given conda an invalid channel. Please double-check
your conda configuration using `conda config --show`.
If the requested url is in fact a valid conda channel, please request that the
channel administrator create `noarch/repodata.json` and associated
`noarch/repodata.json.bz2` files, even if `noarch/repodata.json` is empty.
Traceback (most recent call last):
  File "/io/conda-build-all", line 460, in <module>
    sys.exit(main())
  File "/io/conda-build-all", line 120, in main
    return execute(args, p)
  File "/io/conda-build-all", line 430, in execute
    to_build = list(required_builds(metadatas, pythons, numpys, channel_urls, verbose=args.verbose, force=args.force))
  File "/io/conda-build-all", line 310, in required_builds
    prepend=False)
  File "/anaconda/lib/python3.5/site-packages/conda/core/index.py", line 124, in get_index
    index = fetch_index(channel_priority_map, use_cache=use_cache)
[...]

PR #694, opened 2017-02-05
Don't add conda build dir as channel

jlmaccal:
This might address #690.

I am not sure if this is a valid thing to change, as I don't quite understand the rationale for adding the build dir as a channel in the first place.

We either need to:
1. not add the build dir (as in this PR), or
2. make sure that the build dir has the right structure, to avoid warnings about the lack of a noarch directory and crashes with the latest version of conda.

### DIFF
--- a/conda-build-all
+++ b/conda-build-all
@@ -302,11 +302,8 @@ class VersionIter(object):
 
 def required_builds(metadatas, pythons, numpys, channel_urls, verbose,
                     force=False):
-    if not os.path.isdir(CondaBuildConfig.bldpkgs_dir):
-        os.makedirs(CondaBuildConfig.bldpkgs_dir)
-    conda_build_index.update_index(CondaBuildConfig.bldpkgs_dir, CondaBuildConfig)
     index = conda.api.get_index(
-        channel_urls=[url_path(CondaBuildConfig.croot)] + list(channel_urls),
+        channel_urls=list(channel_urls),
         prepend=False)
     index_by_pkgname = defaultdict(dict)
     for k, v in index.items():


jchodera:
Thanks!

mpharrigan:
The build dir packages are used if you are building multiple packages in one go. E.g. if python 3.6 was released and we rebuilt all packages, the dependencies would be built and saved to that directory and used by the dependers. This is probably moot if we upload each package after building it, because then the internet version would be used and all we'd waste is a little bandwidth

jlmaccal:
@mpharrigan If that's the case, then maybe my PR is going to break builds of PRs and non-master branch, which don't push to anaconda.

I don't quite understand the directory structure that `conda` wants, and I'm not sure why `conda_build.index.udate_index` isn't doing the right thing for us.

jchodera:
Are we *sure* that it's necessary to add the current directory? Doesn't `conda build` stage the built packages in `$MINICONDA/conda-bld/` and automatically use them as needed? It certainly is able to build packages and their dependencies automatically if a recipe and its dependency recipe are in the current directory.

jchodera:
Hm. Now this seems to be forcing rebuilds of dependencies:
```
[conda-build-all] Scheduling the following builds (linux-64):
  pdbfixer-1.4-py27_0.tar.bz2
  pdbfixer-1.4-py34_0.tar.bz2
  pdbfixer-1.4-py35_0.tar.bz2
BUILD START: pdbfixer-1.4-py27_0
Missing dependency openmm, but found recipe directory, so building openmm first
BUILD START: openmm-7.1.0rc1-py27_0
```

jlmaccal:
OK, so we need to figure out how to properly create a local channel repository.

jchodera:
Actually, I wonder if this might be because @peastman just copied 7.1.0rc1 to the `main` label so that the constraints `openmm >=7.1.0` doesn't work as expected. Testing this hypothesis now.

PR #695, opened 2017-02-06
Revert previous changes; add and index noarch

jlmaccal:
This reverts commit f4808280225dd011f137fd86d0c662519f919a63.

Attempts to add a noarch directory that conda expects and creates an index file for it.


jchodera:
This may not be necessary, @jlmaccal! It does appear that builds are working correctly without this. The previous issue I referred to with `pdbfixer` was indeed because the constraint `openmm >=7.1.0` doesn't catch the fact that @peastman just copied over the `7.1.0rc1` package to the `main` label, which comes *before* 7.1.0 in package order.

jchodera:
I think we're good without this, @jlmaccal! Thanks so much for trying to fix things, but we can close this for now (unless you think it needs to be reopened).

PR #698, opened 2017-02-07
Add python 3.6 support (without yet dropping python 3.4 support)

jchodera:
This adds python 3.6 builds without yet dropping python 3.4 support.

### DIFF
--- a/conda-build-all
+++ b/conda-build-all
@@ -95,7 +95,7 @@ def main():
         '--python',
         help="Set the Python version used by conda build",
         metavar="PYTHON_VER",
-        default='27,34,35',
+        default='27,34,35,36',
     )
     p.add_argument(
         '--numpy',


Lnaden:
Isn't this a duplicate of #663?

marscher:
Precisely.

jchodera:
#663 enables python 3.6 support while also dropping python 3.4 support.

jchodera:
It looks like conda-forge *has* dropped python 3.4 support, so it may not be critical to retain this after all. We can discuss on the call today.

jchodera:
Looks like this is probably working on travis, but the log is too long for travis to complete all builds in one go. I'll hold off until after our omnia meeting today to enable this, in case we decide to deprecate python 3.4.

Issue #702, opened 2017-02-07
Python 3.6 compatibility


jchodera:
I'm starting this thread to address some issues that come up in attempting to add python 3.6 support.

The first one, noted [here](https://travis-ci.org/omnia-md/conda-recipes/jobs/199316385#L3804-L3857), is this issue with `msmtools`:
```
BUILD START: msmtools-1.2-np110py36_0
    (actual version deferred until further download or env creation)
Traceback (most recent call last):
  File "/anaconda/lib/python3.5/site-packages/conda_build/build.py", line 1480, in build_tree
    config=config)
  File "/anaconda/lib/python3.5/site-packages/conda_build/build.py", line 945, in build
    create_env(config.build_prefix, specs, config=config)
  File "/anaconda/lib/python3.5/site-packages/conda_build/build.py", line 730, in create_env
    raise
  File "/anaconda/lib/python3.5/contextlib.py", line 77, in __exit__
    self.gen.throw(type, value, traceback)
  File "/anaconda/lib/python3.5/site-packages/conda_build/utils.py", line 523, in path_prepended
    yield
  File "/anaconda/lib/python3.5/site-packages/conda_build/build.py", line 685, in create_env
    warn_on_old_conda_build(index=index)
  File "/anaconda/lib/python3.5/contextlib.py", line 77, in __exit__
    self.gen.throw(type, value, traceback)
  File "/anaconda/lib/python3.5/site-packages/conda_build/utils.py", line 109, in try_acquire_locks
    yield
  File "/anaconda/lib/python3.5/site-packages/conda_build/build.py", line 676, in create_env
    actions = plan.install_actions(prefix, index, specs)
  File "/anaconda/lib/python3.5/site-packages/conda/plan.py", line 487, in install_actions
    update_deps, pinned)
  File "/anaconda/lib/python3.5/site-packages/conda/plan.py", line 662, in get_actions_for_dists
    pkgs = r.install(specs, installed, update_deps=update_deps)
  File "/anaconda/lib/python3.5/site-packages/conda/resolve.py", line 815, in install
    pkgs = self.solve(specs, returnall=returnall)
  File "/anaconda/lib/python3.5/site-packages/conda/resolve.py", line 869, in solve
    self.find_conflicts(specs)
  File "/anaconda/lib/python3.5/site-packages/conda/resolve.py", line 377, in find_conflicts
    raise UnsatisfiableError(bad_deps)
[...]

PR #707, opened 2017-02-17
[conda-build-all] ignore combination py36 and np110

marscher:
Numpy 1.10 is not available (at the moment?) in default channel, so we skip this combination for now. See #702


jchodera:
Thanks!

PR #708, opened 2017-02-17
[conda-build-all] re-enable py36

marscher:


jchodera:
Whoops. Did we jump the gun on the last merge?

marscher:
Maybe. I think this is mergable, the linux build already produced a lot of py36 pkgs.

jchodera:
Thanks!

PR #714, opened 2017-02-23
Add mpi4py and mpich from conda-forge channel to get new mpiexec versions

Lnaden:
We [found a bug](https://github.com/choderalab/yank/issues/347) involving the mpich and CUDA with the version you can get from the default conda channel. In trying to reduce the required channels, we are adding these packages to this channel as copies of the ones in conda-forge.

We can stop building this when we move to conda-forge.


mpich/meta.yaml
```diff
@@ -39,6 +39,7 @@ extra:
     copywrtie:
```
mpharrigan: copywrtie -> copyright?

Lnaden:
I'll have to rerun to get the mpi4py builds on osx, but this should work based on the linux and mpich builds

PR #715, opened 2017-02-23
Fix the tests for scs to work with numpy 1.12

Lnaden:
@rmcgibbo as the one who added the original scs, do these changes look good to you?

### DIFF
--- a/scs/run_test.py
+++ b/scs/run_test.py
@@ -140,13 +140,13 @@ def getConeDims(K):
     l = l + K['ep'] * 3;
     l = l + K['ed'] * 3;
     l = l + len(K['p']) * 3;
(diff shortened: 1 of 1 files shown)


Lnaden:
The windows failures appear to be related to other packages not building on Python3.5/.6. Does anyone see something else in the AppVeyor errors that would cause concern?

rmcgibbo:
Yes, this looks good to me.

Lnaden:
The SCS builds are passing. If the other errors are acceptable to others, this can be merged since these changes only affect SCS.

Lnaden:
The only tests failing are for not scs, merging this so we can keep debugging other builds.

PR #716, opened 2017-02-24
Don't build sphinxcontrib-* as noarch

jchodera:
The `noarch` packages `sphinxcontrib-*` were causing problems with Python 3.6 builds because, even though they were `noarch`, anaconda foolishly cannot identify them as being available when Python 3.6 is in use. Removing `noarch` for now.


jchodera:
@Lnaden : Can you merge this when it passes?

Issue #712, opened 2017-02-19
Skipping Python-3.6 with old Numpy combination


marscher:
Recently I've added a check within the required_builds generator to skip over this combination. However I was not aware, that the VersionIter does a clever filtering of really required builds by inspecting the build requirements. This will then check the first time for py==36 and np=110, skips and because some packages do not require on numpy, these builds will also not be considered (simply because the iterator filtered all combinations of py36, np>110, because np is not a requirement).

A ugly workaround would be to add numpy as a build requirement, even if it is not needed.

marscher:
I have to ask again what were the specific problems with pure python packages? Eg. progress_reporter used to be a (working) pure package, avoiding all of this workaround hassle.

As far as I remember the only reason told was "there are problems" 👍

PR #720, opened 2017-02-27
Fix failing builds on Py36 when NumPy is not Required

### LINKED ISSUE #719: Some Python 3.6 builds are just not compiling
I can't tell what the causes, but it appears as through some of the builds are simply not being staged for Python 3.6. There is no record of them ever being attempted or even considered in the Travis log files, so there is no error to debug.

Most notably is OpenMM, but there is also things like nose-timer on osx. 

The big one is OpenMM which is holding back a number of the other packages from being built on Python 3.6 as well. Is there possibly something going [in the `conda-build-all` code](https://github.com/omnia-md/conda-recipes/blob/master/conda-build-all#L304-L378) which is iterating through the list of Python and NumPy versions and just missing a combination sometimes?

Lnaden:
Builds that did not have NumPy as a requirement were not compiling due to the workaround for Py34/Np112 and Py36/Np110. This problem was pointed out in the original TODO for the workaround. Package names in the error messages were also not matched which is why I missed the logger outputs saying as much.

This PR corrects this problem by attempting to build with a more favorable NumPy version without redundantly running builds when NumPy is specified. Also matches attempted package name with proper Python version. Once merged, existing Python 3.6 builds will want to re-fire to get the changes into their PR.

Fixes #719, Partial resolution to #712 and #702

### DIFF
--- a/conda-build-all
+++ b/conda-build-all
@@ -347,8 +347,6 @@ def required_builds(metadatas, pythons, numpys, channel_urls, verbose,
                             print("Attempting to build {pkgname} on Python 3.6 and NumPy {numpy}".format(
                                 pkgname=base_bldpkg_path, numpy=numpy))
                         rebuild_paths = True
-                        bldpkg_path = get_bldpkg_path(m, python, numpy)
-                        base_bldpkg_path = os.path.basename(bldpkg_path)
                 if python == "34" and numpy == "1.12":
                     if verbose > 1:
                         print("{pkgname} will not build on Python 3.4 and NumPy 1.12".format(pkgname=base_bldpkg_path))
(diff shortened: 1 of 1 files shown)


jchodera:
Thanks for sorting this out!

Lnaden:
We will still have a few problem packages and we will still need to re-run the build code a few times to rebuild all the dependencies before every package builds, but this should at least make it so they *do* build.

jchodera:
Go ahead and merge if it looks like this is helping.

Lnaden:
I'll be making several follow up PR's probably to fix outstanding issues slowly. I'm also not going to delete this branch yet just in case

PR #722, opened 2017-02-27
Turn off Open babel for python 3.6 until its fixed

Lnaden:
Matching issues is openbabel/openbabel#372. I'm keeping an eye on it.


Lnaden:
This correctly deals with the OpenBabel problem for now. We can rebuild this once the bug is fixed

PR #723, opened 2017-02-27
Manually install LatTeX packages to build OpenMM docs

Lnaden:
Sphinx decided to no longer ship the fncychap LaTeX package starting in 1.5, but still make use of it. So to get it, we have to manually install it. It would probably be good to add this to the Docker image at some point.

### DIFF
--- a/devtools/docker-build.sh
+++ b/devtools/docker-build.sh
@@ -12,7 +12,7 @@ bash Miniconda3-latest-Linux-x86_64.sh -b -p /anaconda
 PATH=/opt/rh/devtoolset-2/root/usr/bin:/anaconda/bin:$PATH
 conda config --add channels omnia
 conda install -yq conda-build jinja2 anaconda-client
-tlmgr install fncychap tabulary
+tlmgr install fncychap tabulary capt-of
 
 /io/conda-build-all -vvv $UPLOAD -- /io/*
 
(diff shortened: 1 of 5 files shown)


Lnaden:
Running list of manually installed LaTeX packages in case more crop up:
* fncychap
* tabulary

Lnaden:
Okay, this if finally building OpenMM, but this PR should be reviewed by a couple people since its rather hack-y and patches the OpenMM 7.1rc1 release's docs to compile correctly.

The main problem is that if we want OpenMM 7.1 on Python 3.6 we have to update Sphinx and make some small changes. We can either make those changes here on the publisher side, or change them in the OpenMM code and cut another build, which I imagine is less favorable.

**Summary of changes**
* Several LaTeX packages are added to the build environment since Sphinx 1.5+ no longer ships them.

**Questions for Review**
* Is this patch for OpenMM's docs okay for cutting the Python 3.6 release of 7.1? We will also want to make a PR for OpenMM directly to bring this change in for the next release
* Is not having the `git` hash match for this build acceptable because of the patch acceptable?
* How required is it to ship missing LaTeX packages that Sphinx no longer requires (related issue #725)?
* Do people approve of the additional LaTeX packages in the build in general?

Tagging @peastman @jchodera @jlmaccal

Lnaden:
Any feed back on this? It is kind of holding up the entire distribution system at the moment

peastman:
It is indeed quite a hack!  But for the moment I suppose it's as reasonable as anything.

Lnaden:
Okay, I'll merge this just to get the build going, we can fix this more elegantly later and push a new build if we need to.

Lnaden:
And thanks for looking at this @peastman

PR #724, opened 2017-02-27
Spot release YANK 0.15.2

Lnaden:
Release to include features for active projects for @davidlmobley and @jchodera


Lnaden:
The YANK things look okay, it will probably be a matter of running rebuild a few times since the drive ran out of space from pulling down files then trying to build

PR #726, opened 2017-02-28
Remove boost since its published on default

### LINKED ISSUE #542: Remove our version of boost and peg things to main version of boost
As noted in #512 , now that boost is available in the main channel we should remove our version and peg things to that.

Lnaden:
Pin to Minimum Python 3.6 version. The OSX version of boost takes too long to compile and times out the build. Since this package is available through the main `anaconda` channel, we should be able to safely stop publishing this ourselves.

As this removes a package, PR review would be appreciated.

Fixes #542


Lnaden:
Merging to get the OSX builds going again.

PR #728, opened 2017-03-02
Disable RDKit on OSX

Lnaden:
Its too slow to build before Travis times out preventing other packages from building.


Lnaden:
There appear to be some RDKit packages building on OSX, I'll keep running build to see if they can all get built

Lnaden:
It built, just took most of the time and had to run 1 travis job per build. No need for this now

PR #739, opened 2017-03-13
Pin conda-build to 2.1.5 until they fix it

Lnaden:
`--yes` and `--use-local` conflict in conda-build 2.1.6, no idea when it will be fixed, so pinning version to keep production moving

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -60,7 +60,7 @@ install:
   - conda config --add channels omnia
   - conda config --add channels salilab
   - conda update -yq --all
-  - conda install -yq conda-build jinja2 anaconda-client
+  - conda install -yq conda-build=2.1.5 jinja2 anaconda-client
   - powershell .\\devtools\\appveyor\\missing-headers.ps1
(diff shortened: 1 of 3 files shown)


marscher:
Thank you for fixing this. I thought we had already pinned the conda-build version? Why has it been removed again?

Lnaden:
We had pinned it before I picked up working on this, but I think it was to avoid an issue with a specific conda build version. However, both `conda` and `conda-build` keep releasing versions so rapidly that break backwards compatibility that it was no longer possible to have that pinned version of `conda-build` and an updated `conda`. So we unpinned it, yet here we are again. 

I don't think this particular bug was actually a problem for this repo since the `conda-build-all` script never executed `conda install --yes --use-local`, but I still went through all the repo's I help maintain and pinned it just in case. They appear to have fixed this particular bug, but its not on a cut release yet, so I'm leaving it pinned for now

jchodera:
cc https://github.com/conda/conda/issues/4848 for posterity.

PR #740, opened 2017-03-13
Added Libnetcdf and netcdf4 to the repo

Lnaden:
Omnia packages which depend on it will downgrade/fall behind the longer we are not on conda-forge.
I have added these 2 packages to help keep some of our packages up to date.

Additionally added some `extra` information on other packages I personally have added to make it clear who the point-of-contact should be (me) for the builds as they exist on omnia.

### DIFF
--- a/coverage/meta.yaml
+++ b/coverage/meta.yaml
@@ -36,6 +36,8 @@ about:
   summary: 'Code coverage measurement for Python'
 
 extra:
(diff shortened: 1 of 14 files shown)


jchodera:
Do we need to do this, or would it be better to focus on switching to the CentOS 6 build image and adding conda-forge as a dependency this week?

Lnaden:
This has been superseded by #761

PR #750, opened 2017-04-02
[openforcefield] 0.0.1 test build

jchodera:


jchodera:
Looks like the openforcefield recipe built OK. Failures are unrelated.

PR #756, opened 2017-04-09
Update Lomap recepie. PyQt4 is needed

nividic:
I updated the PyQt version required and the new Lomap git home page

### DIFF
--- a/lomap/meta.yaml
+++ b/lomap/meta.yaml
@@ -4,7 +4,7 @@ package:
 
 source:
   git_rev: 0.0.0
-  git_url: https://github.com/nividic/Lomap.git
+  git_url: https://github.com/MobleyLab/Lomap.git
 
 
 build:
@@ -18,7 +18,7 @@ requirements:
     - boost ==1.61
     - rdkit
     - matplotlib
-    - pyqt
+    - pyqt ==4.11.4
     - networkx >=1.11
     - graphviz
     - pygraphviz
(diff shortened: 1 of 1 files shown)


nividic:
The Linux and Mac tests are failing because OpenMM is not correctly installed in the test environment. My project does not require OpenMM at all. The Windows tests are failing as well. In my original lomap recipe I disabled windows but someone else has changed my original skip section. Can someone help with these
 problems please?

jchodera:
The windows build is still being skipped, so you can ignore the failures on appveyor.

Your recipe actually fixes the lomap failures, so I'm going to go ahead and merge it. The remaining failures are unrelated. Thanks!

When do you plan to add python 3.x support? We are going to need all omnia packages to comply with that at this stage.

davidlmobley:
We're working towards adding python 3.x support, @jchodera - Gaetano is trying to get it in in his "spare time" in the evening in the near future.

nividic:
I'm trying to port Lomap to python3. Unfortunately there is a problem that I cannot think I can fix. In a particular Rdkit library Chem/Draw there is a dependence on the Python Image Library (PIL http://www.pythonware.com/products/pil/) which does not support python3 yet. I send to the developers an email asking about that. The library is quite important to generate molecule depictions that are used to display and output a .png file of the planned alchemical mutation mapping. The alternative would be to generate just a graphviz .dot file.

davidlmobley:
@nividic - I think the priority is to get Python 3 compatibility in, regardless of whether the visualization is retained. I suggest you just make it output a `.dot` file and create an issue on the LOMAP github to resolve visualizations with Python 3 once PIL is Py3 compatible.

PR #758, opened 2017-04-10
Attempt to fix openmm-example-plugin

jchodera:


jchodera:
@peastman : Any idea what we need to fix to get this to build again? It's now the only thing that is failing in omnia.

peastman:
I didn't know it was failing.  I'll look into it when I have time, but other things are higher priorities right now.

jchodera:
Not a problem! 

We may just temporarily disable this while we figure out the build infrastructure upgrade, but we'll keep the recipe around.

PR #667, opened 2017-01-09
[WIP] Sundry changes to work with omnia-linux-anvil image

mpharrigan:

Issue #717, opened 2017-02-24
Update xcode from 6.4 to 7.3.1


Lnaden:
We should probably update the `osx_image` from `xcode6.4` to at least `xcode7.3` since it looks like [Xcode 6.4 support may disappear from Travis](https://docs.travis-ci.com/user/osx-ci-environment/#Xcode-version) any time (Supposedly they dropped support at the end of January)

It also looks like we can just drop the `osx_image` option from the travis file and it will just default to Xcode 7.3.1

Is there a reason we pin to a specific version of xcode other than for consistent build environment?

Lnaden:
There was a typo on the Travis site, they retired xcode 6.1 on Jan 20th, no plans or timeline right now for xcode6.4. The typo has been fixed. (Summarized from Travis-CI support email I sent)

At some point we will need to deal with this, but for now I'm closing this issue.

Lnaden:
I'd like to revive this issue with OpenMM 7.2 coming out soon and Travis [moving to Xcode 8.3 as the default](https://blog.travis-ci.com/2017-11-21-xcode8-3-default-image-announce). This also means we will likely have to bump our [min OSX version to 10.11](https://docs.travis-ci.com/user/reference/osx/#Xcode-7.3), which would be a 2 year old OS instead of a 4 year old OS.

This would impact anyone deploying on OSX who deploys to a specific minimum version.

Something to think about.

jchodera:
Tagging @peastman here.

The [Travis blog post](https://blog.travis-ci.com/2017-11-21-xcode8-3-default-image-announce) says that XCode 8.3 will be default starting Tuesday, 28 November 2017 (today). 

However, we've already [pinned to XCode 6.4](https://github.com/omnia-md/conda-recipes/blob/master/.travis.yml#L7), and this version is not slated for retirement, so this should not affect our ability to build OS X versions compatible with 10.9 and later at the moment.

After the release of OpenMM 7.2, however, we should rapidly explore the feasibility of migrating to conda-forge.

peastman:
Is there any reason to change?  It sounds like 6.4 is still supported and there's no difficulty in sticking with it.  And I really don't like the idea of dropping support for OS versions that are only two years out of date.

jchodera:
I don't see any reason to change the omnia xcode version at this time.

I'm not sure what a conda-forge version would support, however.

Lnaden:
[...]

PR #761, opened 2017-04-17
Update to Omnia Linux anvil 

Lnaden:
Start the process for updating to the CentOS 6 image.

This is very much a work in progress since we cannot reasonably test all the packages at once. So this initial commit will be to make sure the basics of the new docker image work.

This that need to be done in loose order:
* [x] Ensure the docker image is pulled correctly, including EPEL 6
* [x] Ensure the conda-forge channel and packages are being fetched
* [ ] Update some non-OpenMM dependent packages' build number to see if they can build on the new image (future PRs)
* [ ] Remove packages which are in conda-forge that we dont need in omnia anymore (future PRs)

Replaces and closes #667 

Tagging @jchodera


Lnaden:
This will also fix #717

Lnaden:
I have added files to serve as flags for removal, review, and other notes on various packages. Most of these are FLAG_FOR_REMOVAL files to indicate that the package can probably be removed since the package already exists as part of `conda-forge`. These should be reviewed and discussed though.

Lnaden:
Good news! The new docker image can successfully build OpenMM. I have also confirmed that it is correctly using `clang` as the compiler (which now comes from `conda-forge` and not an addon to the image, its finding and building the OpenMMCUDA targets so the CUDA toolkits are installed correctly, its finding and building the OpenMMOpenCL targets from the AMD SDK so those are also installed. The docs are being built which means the TeXLive kit is installed correctly as well.

This validates the docker image is constructed correctly and packages can be built on it. I think it we are ready to move onto testing other packages, removing my debug lines, start reviewing the packages for removal.

Also it would be good if someone else look at least at the docker file to make sure it gets everything we need.

peastman:
Yay!

Lnaden:
@peastman is there a particular reason we need to keep OpenMM pinned to clang 3.8.1? They have clang 3.9.1 on conda-forge as well so this may be a decent time to reassess the compiler version if there is not some hard restriction.

peastman:
In general I'm fine with using the most recent version, but for releases it's important to know and control what compiler is used.  For example, when we release a patch update to fix bugs, it needs to be built with the same compiler version as the release it's a patch to.  Compiler updates have been known to introduce bugs.

Lnaden:
Okay. I'll keep it pinned to 3.8.1 in the main omnia channel since we are not actually releasing a new version yet. We can reconsider the version for any planned major releases going forward.  Thanks!

Lnaden:
Okay, I have updated the README (could use [some review](https://github.com/omnia-md/conda-recipes/blob/4ea36e8450d0470eb2a6d8e72728c7c641ce1a44/README.md) @jchodera ), removed the debugging line, and set all the packages I think are ready for removal and review.

OpenMM builds on the new image, so I think we are ready to go with the migration to CentOS 6 based and conda-forge. I need to reach out to the CF people yet and figure out how to get things like the AMD SDK and the CUDA tools for actually taking OpenMM to CF, but one step at a time here.

When this PR merges: `conda-forge` will become a requirement for packages built after this PR
----

clusterutils/meta.yaml
```diff
@@ -10,6 +10,7 @@ build:
   entry_points:
     - build_mpirun_configfile = clusterutils.build_mpirun_configfile:main
   skip: True  # [not unix]
+  number: 1 # CHECK LINE
```
jchodera: Since [`number:` defaults to zero](https://conda.io/docs/building/meta-yaml.html#build-number-and-string), this will trigger a new build. Just be sure this is what you want!

README.md
```diff
@@ -8,16 +8,70 @@ omnia-md/conda-recipes
 The recipes here create conda packages for scientific and numerical software components associated with the [`omnia`](http://omnia.md) project.
 The packages built from these recipes are shared with the community on [anaconda.org](https://anaconda.org/omnia).
 
-### Installing packages from omnia
-
-To install a package
+## Migration to conda-forge
+The Omnia project has now started migrating to [`conda-forge`](https://conda-forge.github.io/). We encourage all new
```
jchodera: Not *all* new packages: Only those that do NOT depend on OpenMM.

README.md
```diff
@@ -8,16 +8,70 @@ omnia-md/conda-recipes
 The recipes here create conda packages for scientific and numerical software components associated with the [`omnia`](http://omnia.md) project.
 The packages built from these recipes are shared with the community on [anaconda.org](https://anaconda.org/omnia).
 
-### Installing packages from omnia
-
-To install a package
+## Migration to conda-forge
+The Omnia project has now started migrating to [`conda-forge`](https://conda-forge.github.io/). We encourage all new 
+packages to be developed on `conda-forge` now instead of here, and we will be slowly phasing out packages from being 
+deployed on the `omnia` channel. Don't worry though, the packages that are here are not going anywhere, but the newer
```
jchodera: > the packages that are here are not going anywhere

All packages hosted on the omnia channel will be retained to enable reproducibility.

README.md
```diff
@@ -8,16 +8,70 @@ omnia-md/conda-recipes
...
 The packages built from these recipes are shared with the community on [anaconda.org](https://anaconda.org/omnia).
 
-### Installing packages from omnia
-
-To install a package
+## Migration to conda-forge
+The Omnia project has now started migrating to [`conda-forge`](https://conda-forge.github.io/). We encourage all new 
+packages to be developed on `conda-forge` now instead of here, and we will be slowly phasing out packages from being 
+deployed on the `omnia` channel. Don't worry though, the packages that are here are not going anywhere, but the newer 
+versions will instead appear on `conda-forge`.
```
jchodera: New versions of packages built by the omnia build framework will still appear in the omnia channel! 

I think the way you phrase this is unnecessarily confusing. Yes, packages built by conda-forge will appear in the conda-forge channel, but that is obvious.

README.md
```diff
@@ -8,16 +8,70 @@ omnia-md/conda-recipes
...
+    * This will highly depend on the previous point. If we cannot get the CUDA and AMD packages as part of the build 
+      process, then we will strongly need to reconsider this step.
+1. Move packages which depend on OpenMM to `conda-forge`
+1. Finish the move of all packages to `conda-forge`
+
+### How to migrate to `conda-forge` (for existing packages)
+
+PLACEHOLDER
+
+## Installing packages from omnia
```
jchodera: This should be the FIRST thing in the readme. The *most important information* in the whole document is what people should do in order to still be able to install packages from omnia:
* most important: how should they configure things to install from omnia
* second most important: what has changed from how this previously worked

README.md
```diff
@@ -8,16 +8,70 @@ omnia-md/conda-recipes
...
+
+    Important!
```
jchodera: Use a subheading for `Important!` like `#### Important!` if you don't want it to flow into the rest of the text.

README.md
```diff
@@ -8,16 +8,70 @@ omnia-md/conda-recipes
...
+    We highly recommend using the old style of conda channel priority until omnia is entierly on conda-forge!
```
jchodera: We don't recommend it, we *require* it to make the damn thing work.

README.md
Lnaden: John, when we build on `conda-forge`, its pushed to conda forge. The current framework of building packages here in omnia then pushing them to CF is not the CF model. The only way that would work is if we are copying the builds uploaded to the anaconda cloud from the CF channel to the omnia channel with anaconda's tools.

If we want to build things here in omnia and not copy over, then we will have to maintain both the CF feedstocks, and separate omnia feedstocks which is at best just duplication of work, and at worse divergent builds.

Temporarily, we will keep building packages here, but as they move to CF, they will stop being updated on this channel. We are going to have this awkward period where people will need packages from CF and omnia for omnia packages, but when something moves to CF, then it should no longer depend on *anything* from the omnia channel.

jchodera: > Temporarily, we will keep building packages here, but as they move to CF, they will stop being updated on this channel.

The issue here is just of clarity: We want to make it easy for folks to understand what is going on here, and that version of the text didn't make it clear which build framework it was referring to when it said packages would appear on the conda-forge channel.

Lnaden: Okay, I'll rewrite the readme again and get back to you.

Lnaden:
@jchodera I made another pass at the README file, if it looks good to you, I'll double check the CI errors once more and merge if no new errors are introduced

README.md
```diff
@@ -2,22 +2,92 @@
...
 
-### Installing packages from omnia
+## Installing packages from omnia
 
-To install a package
+To install a package (`mdtraj` for example)
 ```bash
+# Set the channel priority behavior
+conda config --set channel_priority false
+# Add conda-forge as the lower priority (FILO format for priority)
```
jchodera: I don't think this is true anymore, since we're using `channel_priority false`. Maybe combine the two channel add commands with a single
```
conda config --add channels conda-forge omnia
```
?

README.md
```diff
@@ -2,22 +2,92 @@
...
+conda config --add channels omnia
 # Install the 'mdtraj' a package
-$ conda install mdtraj
+conda install mdtraj
 ```
 
+The `channel_priority` behavior we configure gives priority to the highest version of the package over all channels.
+ 
+The default behavior (`channel_priority true`) pulls packages from the highest priority chanel, independent of package 
+version numbers. i.e. an older version package on a higher priority channel will be installed over a newer version
```
jchodera: Remove "i.e." and capitalize the sentence.

README.md
```diff
@@ -2,22 +2,92 @@
...
 
+The `channel_priority` behavior we configure gives priority to the highest version of the package over all channels.
+ 
+The default behavior (`channel_priority true`) pulls packages from the highest priority chanel, independent of package 
+version numbers. i.e. an older version package on a higher priority channel will be installed over a newer version 
+of the same package on a lower priority channel.
+
+### Important!
+
+These packages must use the old style of conda channel priority until omnia is entierly on conda-forge!
```
jchodera: entirely

README.md
```diff
@@ -2,22 +2,92 @@
...
+conda install mdtraj
 ```
 
+The `channel_priority` behavior we configure gives priority to the highest version of the package over all channels.
+ 
+The default behavior (`channel_priority true`) pulls packages from the highest priority chanel, independent of package 
+version numbers. i.e. an older version package on a higher priority channel will be installed over a newer version 
+of the same package on a lower priority channel.
+
+### Important!
```
jchodera: Change "Important!" to "Important: Channel priority behavior must be disabled for omnia to work"

README.md
```diff
@@ -2,22 +2,92 @@
...
+
+These packages must use the old style of conda channel priority until omnia is entierly on conda-forge!
+The instructions below enforce the old channel priority format, please visit 
+[the conda documentations on channel priority](https://conda.io/docs/channels.html) 
+for more information.
+
+## Migration to conda-forge
+The Omnia project has started migrating to [`conda-forge`](https://conda-forge.github.io/). New packages that 
+ do not depend on OpenMM should be developed on `conda-forge` and exiting packages which do not depend on OpenMM
```
jchodera: "exiting" -> "existing"

README.md
```diff
@@ -2,22 +2,92 @@
...
+### Planed Migration Stages
+
+1. (**Current**) Update build image to CentOS 6 from CentOS 5
+
+    The base Docker image for linux builds will be updated to CentOS 6 with its new glibc. The base image is the 
+    `conda-forge` anvil, with some [custom addons](https://hub.docker.com/r/jchodera/omnia-linux-anvil/~/dockerfile/) 
+    to include things like the AMD SDK, TexLive, and CUDA for GPU builds. The updated version will ensure packages 
+    can work on the `conda-forge` platform which is CentOS 6 based.
+    
+1. Stop building packages that appear in `conda-forge`
```
jchodera: "For packages that appear in `conda-forge`, remove the corresponding recipes in `omnia`"

README.md
```diff
@@ -2,22 +2,92 @@
...
+
+    We want to minimize the amount of work we have to do as maintainers. To that end, we will stop building things 
+    which freely appear on `conda-forge` and maintained by someone other than us!
+    For reproducibility purposes, we will keep our previously compiled versions, but they will not longer be updated.  
+     
+1. Begin migration of "nearly-pure Python" and non-OpenMM dependent packages
```
jchodera: Allow recipes that do not depend on OpenMM to migrate from `omnia` to `conda-forge`

README.md
```diff
@@ -2,22 +2,92 @@
...
+
+   * Packages which do not depend on OpenMM and can be run on CPUs only should start migrating over to `conda-forge` 
+   in preparation for the total migration. 
+   * Packages which can compile with just the `conda-forge` linux-anvil should also start migrating.
+   * *We highly encourage devs of individual packages to start migrating now.* 
+   * **Once a package is on conda-forge, it should no longer depend packages from omnia!**
+   
+1. Begin migration of `omnia` dependent (but not OpenMM dependent) packages to `conda-forge`
```
jchodera: I'm not sure this stage is meaningful. We don't want users to migrate recipes that depend on other omnia recipes until their dependencies have migrated over---do we?

README.md
```diff
@@ -2,22 +2,92 @@
...
+
+   * As more of the `omnia` dependencies appear on `conda-forge`, more packages can start moving over.
+   * This stage can 
+
+1. Determine the appropriate way to build packages which require more than the `conda-forge` linux-anvil can provide
+
+   * The `conda-forge` linux-anvil does not support some things such as some LaTeX packages, AMD SDK, and CUDA files. 
+   * We will need to reach out to the conda-forge people to see what the best course of action is
+   
+1. Move OpenMM to `conda-forge`
```
jchodera: "Migrate OpenMM to `conda-forge`"

README.md
```diff
@@ -2,22 +2,92 @@
...
+
+    * This will highly depend on the previous point and will probably have the longest time frame.
```
jchodera: Just say "This requires us to identify the best way to add CUDA and AMD APP SDK developer tools to the `conda-forge` `linux-anvil` docker image. We have some ideas, but no concrete solution yet."

README.md
```diff
@@ -2,22 +2,92 @@
...
+      
+1. Move packages which depend on OpenMM to `conda-forge`
```
jchodera: "Migrate remaining packages to `conda-forge`"

This can be combined with the final "Finish the move"

Lnaden:
This PR is ready to go. This will also fix #740 since we will get files from conda-forge

Lnaden:
@rmcgibbo Do we still need the `salilab` channel for anything on Windows? 

I removed the `salilab` channel from the windows requirements. It was added 2 years ago by you for Modeller, however, the `omnia` channel has the version of `modeller` that `ensembler` is pinned to, so I don't think we need it anymore

jchodera:
I think we can get rid of it.

Lnaden:
I already removed it as one of the commits, but I just wanted to double check to be sure.

Lnaden:
i have a couple more checks I want to run. The PR I made on the dev channel indicates something my be off with the openmm install on mac, so Im going to debug that first

Lnaden:
Okay, I fixed the OSX builds by not overwriting `AppleClang` with `clangdev`. But now there is a new problem.

OSX OpenMM build appears to complete, but not install correctly so the test fails to find the `simtk` package like it was not installed right. The linux builds are fine, but there should not have been a change to the osx side. Has anyone ever seen this issue? [file here](https://api.travis-ci.org/jobs/224379567/log.txt?deansi=true), search for `run_test` to get to the failure since its the raw log.

peastman:
No idea.  I don't see any indications in the log of what the problem is.  It runs the PythonInstall target, and appears to complete successfully:

```
[100%] Built target PythonInstall
```

When it gets to the test, it claims it's about to install the package it just built:

```
TEST START: /Users/travis/anaconda/conda-bld/osx-64/openmm-7.1.0rc1-py35_1.tar.bz2
...
The following NEW packages will be INSTALLED:
...
    openmm:          7.1.0rc1-py35_1  local
```

It doesn't report any errors while installing.  Yet it then reports that OpenMM isn't installed:

```
===== testing package: openmm-7.1.0rc1-py35_1 =====
running run_test.py
  File "/Users/travis/anaconda/conda-bld/openmm_1492787167059/test_tmp/run_test.py", line 7, in <module>
    from simtk import openmm
ImportError: No module named 'simtk'
TESTS FAILED: openmm-7.1.0rc1-py35_1
ERROR: File does not exist: /Users/travis/anaconda/conda-bld/osx-64/openmm-7.1.0rc1-py35_1.tar.bz2
```

Possibly that last line is a clue?  Should that file have been created?

Lnaden:
Everything I see tells me it installed correctly, but it looks like the files just were not copied correctly.

> Possibly that last line is a clue? Should that file have been created?

Sadly no, that is just the generic error that conda-build throws as the last step as it checks for the tarbal as its last action, independent of if something else failed. 

I think this might be conda problem, I have some things I can test to find out.

Lnaden:
I cannot find a way to build on OSX without conda-build just failing for some unknown reason. Moving the channel priority to later appears to break the `conda-build-all` script. There are no helpful errors. Will keep debugging

Lnaden:
So the problem I am seeing on the OSX build is exactly the same problem we are seeing on the Python 3.6 windows builds here: https://github.com/pandegroup/openmm/issues/1758#issuecomment-289535785

How to fix it though....

Lnaden:
Okay, here is what I can tell.....

The problem comes from the fact that python is built against a certain `MACOSX_DEPLOYMENT_TARGET` minimum which is separate from the environment variable. This appears to be tied to the python executable itself from what I can read (I think, there may be a way to change this). 

When building new packages, the environment `MACOSX_DEPLOYMENT_TARGET` cannot be lower than the python that was used to compile it. So what we are seeing is the python we get on OSX for python versions not 3.4 are built against osx 10.9, then when we set the env `MACOSX_DEPLOYMENT_TARGET`, it complains and crashes.

There are some oddities with this though. My local version of Python 3.5.3 from `default` channel was built against macosx 10.7, but the same version of python 3.5.3 from `conda-forge` is built against osx 10.9. My guess is the version of python 3.4, which stopped getting updates, was built against the 10.6 osx, which is why we see this difference. 

So now the question is, can we set the version to built against from the python compiler, or are we going to have to find that magic python version which is 10.7 compiled for each version of python.

On a side note, this is excessively dumb.

jchodera:
> On a side note, this is excessively dumb.

+1000

jchodera:
Maybe this is a good time to report it on the `conda-forge` and possibly `anaconda` issue trackers to solicit advice?

Lnaden:
> Maybe this is a good time to report it on the conda-forge and possibly anaconda issue trackers to solicit advice?

Maybe, the problem is this does not appear to be a `conda`-related issue. I think this is a pure-python issue since I think `setup.py` scripts would throw this issue. Evidence [here](https://bugs.python.org/issue9516) where the Python folk were pointing it out back in Python 3.2/3.3. This is my evidence that it looks like a Python deployed to a min version cannot deploy packages it builds to an even smaller min version.

jchodera:
It's conda-related insofar as the `python` versions made available in these channels were compiled with specific versions of the OSX frameworks, right?

Are all the `conda-forge` python versions built with 10.7 or earlier? If so, could we just
* add the `conda-forge` channel
* `conda clean -pltis --yes` to get rid of cached anaconda python installs
* create a new env with the `conda-forge` python
* then add other channels and build?

Lnaden:
I just realized this conversation is happening on the wrong issue, linking back to the comment [here](https://github.com/omnia-md/conda-dev-recipes/pull/86#issuecomment-296695820)

Lnaden:
Hold off on merging this until OpenMM 7.1.1 is published at minimum

I think this is ready to go for the most part (some last builds running)

jchodera:
Should we at least disable the current failing byulds? I'm a separate pr?

Lnaden:
I can and did for most of them already. The remaining errors being raised are legitimate concerns that I don't want to silence. They are mostly caused by lack of OpenMM on Python 3.6 for Windows that we never fixed. When we cut 7.1.1, we will want to fix that first.

Lnaden:
And to be precise, the remaining errors are *all* related to missing OpenMM builds.

Lnaden:
Guess what! Its merging day!

PR #762, opened 2017-04-17
Remove parmed dependence on openmm

jlmaccal:
Since openmm is an optional dependency of parmed, can we remove the dependency?

We are deploying our own versions of openmm and meld built against specific cuda versions. Installing parmed from the omnia channel tries to pull in openmm, which leads to a conflict.

### DIFF
--- a/parmed/meta.yaml
+++ b/parmed/meta.yaml
@@ -7,7 +7,7 @@ source:
   url: https://github.com/ParmEd/ParmEd/archive/2.7.3.tar.gz
   
 build:
-  number: 0
+  number: 1
 
 requirements:
   build:
--- a/parmed/meta.yaml
+++ b/parmed/meta.yaml
@@ -19,7 +19,6 @@ requirements:
     - python
     - numpy
     - pandas
-    - openmm
 
 about:
   home: http://parmed.github.io/ParmEd


swails:
Makes sense to me. But I'm surprised conda doesn't see that the openmm dependency is already satisfied when the simtk package is present. 

Maybe if the importable package was openmm it would work?

jlmaccal:
The issue is that my lab has `meld-cuda75`, `openmm-cuda75`, etc packages. These are distinct from the `openmm` packages that omnia provides, so conda will happily install both. However, they have the same paths and will overwrite each other.

Currently, we have to do something like:
```
conda install -c omnia parmed
conda remove --force openmm
conda install -c maccallumlab meld
```

This isn't to bad, but it's a total disaster to then update any of the packages later. I'm currently looking into conda variant packages to see if we can find a better way to deal with different cuda versions.

jlmaccal:
Can someone merge this?

jchodera:
@jlmaccal : Should we also bump the build number to force the build of a new package? Otherwise, no new package would be built.

jlmaccal:
OK, bumped the build number.

jchodera:
Thanks!

PR #764, opened 2017-04-18
Update dockerfile to update yum repository

jlmaccal:
This adds commands to update the yum repository to point to the correct new location and installs `groff`, which is needed for `awscli`.

@jchodera or @lnaden, can you trigger a new build of the docker image?

### DIFF
--- a/devtools/centos5-build-box/Dockerfile
+++ b/devtools/centos5-build-box/Dockerfile
@@ -13,6 +13,11 @@ ADD texlive.profile .
 # The other TeX packages installed with `tlmgr install` are required for OpenMM's sphinx docs
 # libXext libSM libXrender are required for matplotlib to work
 
+# Update yum repositories to point to vault.centos.org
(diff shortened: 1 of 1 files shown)


jchodera:
No, that's the wrong Dockerfile!

@Lnaden can add them to the new `omnia-linux-anvil` if you just let him know what you need.

@Lnaden : We should also remove the deprecated dockerfile in `devtools/` so it doesn't cause confusion.

jchodera:
Sorry about this, @jlmaccal. @Lnaden *is* actively working on this---he spent all day working on the update.

jlmaccal:
Oops.. in that case only
```
RUN yum install -y groff
```
is the only necessary addition. The repository changes were only necessary for CentOS5.

Lnaden:
>  We should also remove the deprecated dockerfile in 

Yup, already part of the PR.

@jlmaccal The docker file we will be switching to is at [this repo](https://github.com/omnia-md/omnia-linux-anvil) which auto-rebuilds the docker image that the new builds will use. I'll take care of this first thing in the morning and for now will re-open this ticket/PR to serve as a reminder to myself.

PR #765, opened 2017-04-21
[WIP] New pygraphviz recipe

nividic:
Try to build pygraphviz

### DIFF
--- a/pygraphviz/build.sh
+++ b/pygraphviz/build.sh
@@ -1,2 +1,10 @@
-#!/bin/bash
-C_INCLUDE_PATH=$PREFIX/include $PYTHON setup.py install
+#!/usr/bin/env bash
+
+set -x -e
+
+export LIBRARY_PATH="${PREFIX}/lib"
+export INCLUDE_PATH="${PREFIX}/include"
+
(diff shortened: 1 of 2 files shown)


Lnaden:
I would hold off on merging this for just a bit (this morning, 7/21/17), I am about to merge the #761 which will change the docker build image. We will require `conda-forge` as a channel after that PR along with `omnia`, so the install instructions will change.

We also encourage migrating packages over to `conda-forge` after that as well, so it might be worth trying to build the package there instead.

Lnaden:
This PR is having issues building, missing things like `coverage` and `toolchain` (which I think is conda-forge only). We won't be able to merge the PR which makes this conda-forge ready (#761) until OpenMM patch 7.1.1 is ready. I don't have an exact timeline, but it should not be too long. We have merged the dev version of the conda-forge PR and so far things look good.

If you can fix this PR without conda-forge requirements before we get the 7.1.1 patch for OpenMM in, feel free to merge this.

The toolchain from conda forge is basically just a bunch of sym links and environment variables, so it should be possible to build without it, but then the `build.sh` script will be different.

PR #767, opened 2017-04-27
msmbuilder 3.8.0

brookehus:
new release


jchodera:
MSMBuilder successfully built! I'll merge this now.

Note that the build system upgrade is being held on @peastman's request for the OpenMM 7.1.1 bugfix release, so it'll be a few days before those travis failures are cleared.

PR #768, opened 2017-05-01
[openmm] 7.1.1

jchodera:
This should build the 7.1.1 release candidate and push it to the `rc` label.


Lnaden:
So OpenMM is building when Travis does not time out (nvidia problems). However, there are some problems with it. 

1. The linux builds throw an error on doc creation with this error: `gmake[4]: warning: jobserver unavailable: using -j1.  Add `+' to parent make rule`. I can't directly link it since its only in the raw log. The build completes, but I'm concerned the docs are not building.
2. The CPU energies are not computing throwing this 
```
2 CPU - Error computing forces with CPU platform
CPU platform error: St8bad_cast
```
Normally the CPU computes fine. The builds from the main OpenMM site and from the conda-dev recipes which are already on conda-forge work fine too. So I'm not really sure what is happening here.

*Caveat*: The dev channel and the main site have all the changes, not just the cherry picked ones for 7.1.1

@peastman have you ever seen this problems?

jchodera:
Is `$CPU_COUNT` not being defined?

We can change lines like these to get rid of the `-j#` argument if this is causing problems:
https://github.com/omnia-md/conda-recipes/blob/96dafcabbfc4d5644e89c8927a9a7286c3679dfa/openmm/build.sh#L114

peastman:
> CPU platform error: St8bad_cast

I haven't seen anything like that.  It works when I run it locally.

jchodera:
We've seen this error in `conda-dev-recipes` at least as far back as Mar:
https://travis-ci.org/omnia-md/conda-dev-recipes/jobs/208704834#L6967

Perhaps there is some weird system on travis for limiting the number of threads?
I bet if we set the environment variable `OPENMM_CPU_THREADS=1` on the same line that the test is run it would work fine.

peastman:
Do you have any way to reproduce that error locally?  It's hard to even guess where it's happening or what could be causing it.

jchodera:
I think you can [run travis docker images locally](https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally), but I'm not sure if that would exactly reproduce the issue since the underlying hardware will be different.

jchodera:
Welp, I was wrong---`OPENMM_CPU_THREADS=1` didn't help.
```
===== openmm-7.1.1-py27_0 OK =====
+ /bin/bash -x -e /anaconda/conda-bld/openmm_1493686863494/test_tmp/run_test.sh
+ env OPENMM_CPU_THREADS=1 python -m simtk.testInstallation
There are 4 Platforms available:

1 Reference - Successfully computed forces
2 CPU - Error computing forces with CPU platform
NVIDIA: no NVIDIA devices found
NVIDIA: no NVIDIA devices found
NVIDIA: no NVIDIA devices found
3 CUDA - Error computing forces with CUDA platform
4 OpenCL - Error computing forces with OpenCL platform

OpenCL platform error: Error initializing context: clGetPlatformIDs (-1001)

CPU platform error: St8bad_cast

CUDA platform error: Error initializing CUDA: CUDA_ERROR_UNKNOWN (999) at /anaconda/conda-bld/openmm_1493686863494/work/platforms/cuda/src/CudaContext.cpp:148
```

peastman:
Are we certain this only affects the CPU platform?  It looks possible that the OpenCL and CUDA platforms would also have the same problem, only they fail before they can reach it due to the lack of an available device.

The way you get a `bad_cast` is by doing a `dynamic_cast` on a reference, and we don't do that in many places.  It looks to me like this is probably happening when it tries to cast a KernelImpl defined in the plugin, such as casting a `CpuCalcNonbondedForceKernel&` to a `CalcNonbondedForceKernel&`.  And it somehow thinks it's not actually an instance of that.  Perhaps this could be caused by using different compilers or different compilation flags for the plugin and the main library?  But I don't see why that would be happening.  It isn't trying to build static libraries, is it?

Lnaden:
I'm not sure that its limited to just the CPU platform, but there is no way to test it on CUDA and OpenCL on travis. 

I can't find anything that changed between now and the last time 7.1.0rc1 was built which was back in December. The only thing that should have changed was the 7.1.1 code, and maybe the conda/conda package versions. The build image did not change.

peastman:
There haven't been any changes to the CPU platform, or any changes to the build scripts.  So that doesn't seem like an explanation.

John, you said it's been happening on conda-dev-recipes for a while.  Do you know when it started?

jchodera:
> John, you said it's been happening on conda-dev-recipes for a while. Do you know when it started?

I'm at a conference today, but maybe @Lnaden can look through the travis history for that repo?

Lnaden:
Can do!

Lnaden:
The CPU throwing `bad_cast` has been a problem since the 7.1_branch of OpenMM was created *at least*.

From the logs in travis, the CPU platform has been failing with the `bad_cast` since [88b0a3e](https://github.com/pandegroup/openmm/commit/88b0a3e) at the earliest and the 7.1_branch builds on travis have been generating `bad_cast` since its creation. I can't look back further than build [472 on Travis](https://travis-ci.org/omnia-md/conda-dev-recipes/jobs/184657609) as the logs dont load past that point; probably not saved, its >4 months old at this point.

I also don't think that the `jobserver unavailable` errors are fatal as it appears to not stop the build.

peastman:
> The CPU throwing bad_cast has been a problem since the 7.1_branch of OpenMM was created at least.

I just created that branch a couple of days ago.  You can't tell anything before that?

Lnaden:
Sorry. Got my branches mixed up, its been doing that since the creation of the `openmm-7.1-beta` on the `conda-recipes` repo which was 5 months ago.

Lnaden:
...Which is when it was pinned to commit 88b0a3e on OpenMM

Lnaden:
I would guess that it goes back further since the only changes made on that commit were Windows changes and the issue is in Linux. However, the Travis logs do not appear to be saved back that far, or at least are throwing errors.

Lnaden:
It may be worth noting the problem goes away after I merged in the changes to make the `dev-recipes` channel switch to conda-forge.

peastman:
> its been doing that since the creation of the openmm-7.1-beta on the conda-recipes repo which was 5 months ago.

Do you mean conda-recipes or conda-dev-recipes?  The former clearly was working more recently, since we didn't have this problem when we built 7.1.  But if conda-dev-recipes was failing while conda-recipes was working, that certainly suggests the problem is in the recipe.

Lnaden:
I specifically meant conda-dev-recipes, thats the only history I went over. It will be a bit harder to track down if the build ever failed on conda-recipes since I will need to go find the last time OpenMM actually built on that repo. I suppose I can go look at the history and try to find the CI checks from that PR.

peastman:
> I specifically meant conda-dev-recipes

That's what I assumed.  You said conda-recipes, which is why I was confused.

So this error has been happening on conda-dev-recipes for a while, but just started happening on conda-recipes recently (some time since 7.1 was built).  Of the various updates since then, this one looks the most suspicious: https://github.com/omnia-md/conda-recipes/commit/528c7ca2f2d82b30e4e9d035af361bd8c7a7fb6d.  It was checked in the day after 7.1 was released.  It has the innocuous sounding description, "Rename to 7.1.0rc1," but in fact makes tons of changes to lots of files.

Lnaden:
I found that too, but I am trying to find exactly which change introduced the error. The problem is that the CI tests often never make it to the meta.yaml test section because of an earlier crash

Lnaden:
So you are right in that it is that change to the main recipe when the errors started showing up. However, the change before that was back on #534 where it was still working. 

Now trying to find exactly *which* change to the file caused this to fail, I have looked through the conda-dev-recipes version of the meta.yaml file and the change appears to happen on [this version](https://github.com/omnia-md/conda-dev-recipes/blob/3e53537ac5775af1fca801af47db02ded8425eb1/openmm/meta.yaml) of the file with [this diff which changed from triclinicerror tag to 88b0a3e](https://github.com/omnia-md/conda-dev-recipes/commit/3e53537ac5775af1fca801af47db02ded8425eb1). At that point, the error started appearing. It was already on swig 3.0.8, and there were no other changes to the file. So I'm not sure the issue comes from a large change to the meta.yaml file.

Lnaden:
Although.... I can't seem to access the travis logs now. ugh, maybe im wrong

peastman:
Any progress on figuring this out?

Lnaden:
Not really, I can't reproduce locally, I can't even get the docker image working right. This has been a symptom for over 5 months now though and I have not heard users complain, so it may be related to Travis-CI or something that changed running the linux build in a docker image on travis unable to access the CPU.

I say we fix the git hash to get the addition of that one other issue, rebuild, merge onto dev, and test from the dev tag

peastman:
Sounds like a plan.  Do you need me to do anything?

Lnaden:
The Windows build. For the Linux and OSX ones, probably not.

Lnaden:
OSX and Linux builds are completing for OpenMM, merging so we can start local testing

Lnaden:
OSX and Linux builds are up on the rc tag, ready for testing. 

@peastman Do you need anything from us for helping with the Windows builds?

peastman:
I managed to build the files for the Windows zip installer.  I don't know why it failed before.  I updated the code to get the one extra change, uninstalled FFTW, reinstalled it, ran the build script, and this time it worked.

Which recipe should I use for building the conda installer?  The one here, or the one at conda-dev-recipes?

Lnaden:
Do you mean the `conda-recipe` For the 7.1.1 files? The one here

PR #769, opened 2017-05-03
gpflow

cxhernandez:
Add [GPFlow](https://github.com/GPflow/GPflow/) to eventually replace GPy

### DIFF
--- a/gpflow/bld.bat
+++ b/gpflow/bld.bat
@@ -1,0 +1,2 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
--- a/gpflow/build.sh
+++ b/gpflow/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+python setup.py install
--- a/gpflow/meta.yaml
+++ b/gpflow/meta.yaml
@@ -1,13 +1,13 @@
 package:
   name: gpflow
-  version: 0.3.5
+  version: 0.3.5a
 
 source:
   git_url: https://github.com/GPflow/GPflow.git
-  git_tag: 0.3.5
+  git_tag: 103e88122a185069312426dab5b5d61669211b7f
 
 build:
-  skip: True  # [win or osx]
+  skip: True  # [win or osx or py34]
 
 extra:
   maintainers:
--- a/gpflow/meta.yaml
+++ b/gpflow/meta.yaml
@@ -1,0 +1,37 @@
+package:
+  name: gpflow
+  version: 0.3.5
+
+source:
+  git_url: https://github.com/GPflow/GPflow.git
+  git_tag: 0.3.5
+
+build:
+  skip: True  # [win or osx]
+
+extra:
+  maintainers:
+   - cxhernandez
+
+requirements:
+  build:
+    - python
+    - scipy
+    - numpy >1.10
+    - pandas
+    - tensorflow
+
+  run:
+    - python
+    - scipy
+    - numpy >1.10
+    - pandas
+    - tensorflow
+
+test:
+  imports:
+    - GPflow
+
+about:
+    home: https://github.com/GPflow/GPflow/
+    license: ALv2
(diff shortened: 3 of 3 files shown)


cxhernandez:
@omnia-md/members: Is there a way to skip numpy 1.10 builds? I tried adding `np == 1.10` to the skip builds section, but it doesn't seem to be working.

mpharrigan:
`110`
https://conda.io/docs/building/meta-yaml.html#preprocessing-selectors

cxhernandez:
Thanks!

cxhernandez:
Nothing seems to be building in the Linux test 😕

Lnaden:
I know half of what is wrong, and its a bug, but there is a fix you can do to your `meta.yaml` file to work around it (and I think you need it anyways).

1. Change your `skip: True  # [win or np == 110]` to `skip: True  # [win]`
1. Add the following to your file in the `requirements` section: 
  ```
  run:
    - python
    - scipy
    - numpy >1.10
    - pandas
    - tensorflow
``` 
  * This is missing so your file is saying it does not need these things to run, only to build. The build would be deployed and people could then install this package without python technically.
3. Change the `numpy` in the `build` section to be the same as the run section I showed.

This will require people have numpy > 1.10 if they want to use your package. Now if you want people to specify which numpy they use, you would need `numpy x.x`, and then I would also have to fix the `conda-build-all` script for a few bugs to get that to work right again.

`conda-build` and the `meta.yaml` format has changed so much from when our build script was written, things like this crop up. We are trying to move to the `conda-forge` framework partially so we dont have to maintain our own version of this script. If you need the `numpy x.x` function, I will need to spend some time fixing the script, but I don't know when I will have the time to devote to that in the near future right now.

**Comment by mpharrigan** [comment]
does gpflow compile against the numpy api?

**Comment by cxhernandez** [comment]
> does gpflow compile against the numpy api?

No but it's a tensorflow requirement, which currently only works with numpy >1.10

**Comment by cxhernandez** [comment]
Thanks @Lnaden! Try that now

**Comment by mpharrigan** [comment]
skipping numpy builds only makes sense if you're compiling different packages for different numpy abi's (as @Lnaden has pointed out)

**Comment by cxhernandez** [comment]
Actually nvm:

```
install_requires=['numpy>=1.9', 'scipy>=0.16', 'pandas>=0.18.1'],
```

Looks like it [does](https://github.com/GPflow/GPflow/blob/master/setup.py#L49)

**Comment by Lnaden** [comment]
The requirements in the `setup.py` file are roughly equivalent to the conda `meta.yaml` requirements' `run` section if I understand them right. So long as you don't really have a conflict, it should build as I put it

**Comment by cxhernandez** [comment]
Ahah! That seems to have gotten the builds started. Now for a new issue:

```
ImportError: /lib64/libc.so.6: version `GLIBC_2.9' not found (required by /anaconda/conda-bld/gpflow_1493845345443/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.5/site-packages/tensorflow/python/_pywrap_tensorflow.so)
Failed to load the native TensorFlow runtime.
```

Lnaden:
One other thing, there is no osx tensorflow on the anaconda default channel, so these builds will not work on OSX.

mpharrigan:
Looks like tensorflow needs centos > 5. @Lnaden has the omnia migration to centos 6 ready but we're waiting to build an openmm patch release before it's merged.

cxhernandez:
Cool! Seems like Linux builds just fine now 😄

Lnaden:
This will have to wait until we can merge in the conda-forge PR.

cxhernandez:
That's fine. I just wanted to be sure it would work, since this will be the dependency for something currently in development.

PR #786, opened 2017-06-05
-y option no longer valid in homebrew-cask?

Lnaden:
Not sure what is wrong here, but something changed either in Homebrew or Homebrew-cask that made it so the -y option isnot accepted in `brew cask` anymore. I don't know if this is a bug or intentional yet and no one has raised an issue on either github, so I am writting in a || to try and error trap a bit for both cases.

### DIFF
--- a/devtools/osx-build.sh
+++ b/devtools/osx-build.sh
@@ -31,7 +31,7 @@ if [ "$INSTALL_OPENMM_PREREQUISITES" = true ] ; then
     rm -f cuda_mac_installer_tk.tar.gz cuda_mac_installer_drv.tar.gz
 
     # Install latex.
-    brew cask install -y basictex
+    brew cask install -y basictex || brew cask install basictex
     export PATH="/usr/texbin:${PATH}:/usr/bin"
     sudo tlmgr update --self
     sleep 5
--- a/devtools/osx-build.sh
+++ b/devtools/osx-build.sh
@@ -31,8 +31,17 @@ if [ "$INSTALL_OPENMM_PREREQUISITES" = true ] ; then
     rm -f cuda_mac_installer_tk.tar.gz cuda_mac_installer_drv.tar.gz
 
     # Install latex.
-    brew cask install -y basictex || brew cask install basictex
     export PATH="/usr/texbin:${PATH}:/usr/bin"
+    if brew cask install -y basictex
+    then
+        # Do nothing if the command completed
+        true
+    else
+        # Trap the error and install using new version
+        brew cask install basictex
+        # Path based on https://github.com/caskroom/homebrew-cask/blob/master/Casks/basictex.rb location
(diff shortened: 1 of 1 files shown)


Lnaden:
Anyone with a better grasp on Homebrew know why this is now failing to find the `tlmgr` from basictex?

```
+brew cask install basictex
==> Downloading http://mirror.ctan.org/systems/mac/mactex/mactex-basictex-20170523.pkg
==> Verifying checksum for Cask basictex
==> Installing Cask basictex
==> Running installer for basictex; your password may be necessary.
==> Package installers may write to any location; options such as --appdir are ignored.
==> installer: Package name is BasicTeX-2017
==> installer: Installing at base path /
==> installer: The install was successful.
🍺  basictex was successfully installed!
+export PATH=/usr/texbin:/Users/travis/anaconda/bin:/Users/travis/.rvm/gems/ruby-2.0.0-p643/bin:/Users/travis/.rvm/gems/ruby-2.0.0-p643@global/bin:/Users/travis/.rvm/rubies/ruby-2.0.0-p643/bin:/Users/travis/.rvm/bin:/Users/travis/bin:/Users/travis/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/bin
+PATH=/usr/texbin:/Users/travis/anaconda/bin:/Users/travis/.rvm/gems/ruby-2.0.0-p643/bin:/Users/travis/.rvm/gems/ruby-2.0.0-p643@global/bin:/Users/travis/.rvm/rubies/ruby-2.0.0-p643/bin:/Users/travis/.rvm/bin:/Users/travis/bin:/Users/travis/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/bin
+sudo tlmgr update --self
sudo: tlmgr: command not found
```

jchodera:
It looks like `tlmgr` is either (1) not on the `$PATH`, or (2) not installed by the `basictex` package anymore. 

It *looks* like the [cask recipe](https://github.com/caskroom/homebrew-cask/blob/master/Casks/basictex.rb) was just updated on May 23:
```
  version '20170523'
```
You might see if we could force an older version to install.

Lnaden:
The problem is that it was working 4 days ago, and the basictex recipe did not change. I think it is actually a homebrew or homebrew-cask change, but I've got an idea to work around it.

Lnaden:
Okay, so it looks like this will go through finally. It feels really hack-y, but I don't see an easier fix. Does this work-around look like its a decent solution? It *should* still work as the basictex version gets updated from year to year, unless they change the install path, or homebrew(-cask) does something else to break what we had working.

Lnaden:
Failing due to time outs of builds, but the problem this PR fixes is addressed. I'm going to merge this to get the builds running again, we can fix anything I broke at a later time

PR #787, opened 2017-06-20
[assaytools] 0.3.0

jchodera:

PR #790, opened 2017-06-21
Pin to texlive repository

### LINKED ISSUE #789: Force tlmgr to use a non-broken mirror

Lnaden:
This tries to pin to a specific repository to avoid getting an error because the CTAN repo is out of date.

Fixes #789

### DIFF
--- a/devtools/osx-build.sh
+++ b/devtools/osx-build.sh
@@ -32,22 +32,16 @@ if [ "$INSTALL_OPENMM_PREREQUISITES" = true ] ; then
 
     # Install latex.
     export PATH="/usr/texbin:${PATH}:/usr/bin"
-    if brew cask install -y basictex
-    then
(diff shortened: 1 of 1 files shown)


jchodera:
That's weird: Assaytools is building on appveyor, but not importing correctly:
https://ci.appveyor.com/project/jchodera/conda-recipes/build/1.0.271/job/1sukik1fe3r5gtqj#L178

Lnaden:
@jchodera it looks like the assay tools is failing as there are no files which are copied. They appear to be unrelated to this PR though as #787 was not passing its Windows CI either.

I'm going to merge this though to get the OSX builds stable at minimum

jchodera:
I think we have to do the same thing to https://github.com/omnia-md/conda-dev-recipes as well.

PR #793, opened 2017-06-30
[openmmforcefields] 0.1.0 preview release

jchodera:
This adds conda-installable AMBER forcefields for OpenMM.


jchodera:
This package requires `openmm >= 7.2.0`, so I am building it in [`conda-dev-recipes`](https://github.com/omnia-md/conda-dev-recipes/pull/95) instead until we release the new OpenMM.

PR #796, opened 2017-07-07
Hotfix Bump OpenMMTools to 0.11.2

Lnaden:


Lnaden:
Merging this as the failure appears to be some new python/conda bug in windows 2.7, not sure why, but unrelated to this hotfix

PR #798, opened 2017-07-07
Pin Windows to Conda-build 2.1.17

### LINKED ISSUE #797: Conda-Build Fails on Win Py27
Windows Python 2.7 builds have started failing due to string + list concatenation failures, for... reasons!

Looks like Conda-build was updated on that platform + py version from 2.1.17 to 3.0.3. So API broke somewhere probably. 

This issue is to document and serve as a reminder to fix.

Lnaden:
Fixes #797
More of a work-around, but doing a full upgrade of conda-build may require much more work


Lnaden:
Window's build is passing now. Merging this to get the windows builds going again.

PR #802, opened 2017-07-19
Bumping DeepChem to 1.2.0

rbharath:
Updating version to latest DeepChem.

### DIFF
--- a/deepchem/meta.yaml
+++ b/deepchem/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: deepchem
-  version: "1.1.0"
+  version: "1.2.0"
 
 source:
     git_url: https://github.com/deepchem/deepchem.git
-    git_tag: 1.1.0
+    git_tag: 1.2.0
 
 build:
   number: 0


rbharath:
I'm seeing some mysterious errors on Travis. Does anybody have any thoughts on what's going on?

Lnaden:
Travis recently switched from Precise to Trusty as their default image, I'm guessing the update of `libc` is causing the problem. As to a fix... You may need to go look through the issues in the [tensorflow conda-forge feedstock](https://github.com/conda-forge/tensorflow-feedstock) to see if someone has looked into this problem. Their linux builds go through CircleCI as I understand though, so they may not know about this problem yet.

I could of course, be totally wrong on what is happening though.

lilleswing:
@Lnaden if you look carefully everything is actually getting built on a docker image 

``` bash
        docker run -e UPLOAD -e BINSTAR_TOKEN -e TRAVIS_PULL_REQUEST
            -t -i --rm -v `pwd`:/io jchodera/omnia-linux-anvil:texlive-amd30-cuda${CUDA_SHORT_VERSION}
            bash /io/devtools/docker-build.sh;
```

This docker image is running centos 6.9 with libc 2.12
``` bash
(root) [root@f896f82eea3c /]# cat /etc/centos-release
CentOS release 6.9 (Final)
(root) [root@f896f82eea3c /]# ldd --version
ldd (GNU libc) 2.12
```

CentOS 9 stopped getting full support in May 2017 -- we should switch the anvil over to CentOS 7 which has a newer glibc (Although I understand that might be a very large undertaking).  I don't see how we can make this work with the current setup.

Lnaden:
Right, the libc in the docker image would be the one used, not the travis image.

We actually will continue using the CentOS6 since our docker image is actually an extension of the `conda-forge` one, one that I personally spent quite a while getting up and running. As we are trying to move packages to conda-forge anyways, we want to stay as in-sync with them as we can. Also, Cent 6 EOL wont be until Nov 2020 anyways, so we have time that it will still be supported.

The part I find odd is that the `tensorflow` package is coming from conda forge, where it successfully built on their docker image, of which ours is an extension. As such, I don't see how `libc` is actually causing a problem, which is why I am very confused as to what is happening here.

mpharrigan:
https://github.com/conda-forge/tensorflow-feedstock/issues/6

mpharrigan:
Maybe you could try installing tensorflow from the defaults channel? Perhaps the anaconda people have solved this problem

jchodera:
> conda-forge/tensorflow-feedstock#6

From that issue: "[This comment](https://github.com/scipy/scipy.org/pull/179#issuecomment-247179912) says that indeed tensorflow pulls in a numpy wheel somehow. So that's pretty badly broken."

Ouch. Maybe you can give the feedstock maintainers a hand sorting that mess out, @lilleswing?

PR #807, opened 2017-08-21
Bump Yank to 0.17.0

Lnaden:


Lnaden:
The error from here is due to Travis OSX timeout, not because tests fail. Because there are a few OSX builds which do succeed, this should be resolved with a few runs through the osx builds. Merging

PR #808, opened 2017-08-24
HMMLearn v0.2.0

cxhernandez:


jchodera:
For some reason, one of the python 3.6 win builds hung, but I don't think that's related. 
We can merge this PR if it's ready to go!

cxhernandez:
Finally!  After adjusting so many different things!

PR #810, opened 2017-08-31
Bump shiftx2 to v1.10

mpharrigan:


jchodera:
Since `shiftx` doesn't depend on `openmm` and `mdtraj` is migrating to `conda-forge`, what about moving this recipe to `conda-forge`?

mpharrigan:
mdtraj's *tests* still depend on openmm, so our CI will rely on the omnia channel for the foreseeable future. I'm just trying to fix a bizarre issue that shows up only on travis (https://github.com/mdtraj/mdtraj/pull/1292) with shiftx2

In general, I'm not well versed in chemical shift prediction and I don't want to be the maintainer of the shiftx recipe on conda-forge

mpharrigan:
the osx build timed out for an unknown reason. but this is linux only, so ok to merge?

jchodera:
Oh, sure!

mpharrigan:
gah this didn't actually build anything. probably because it can't figure out 1.10 > 1.9A due to lack of semver

mpharrigan:
```
Evaluating whether to build package /opt/conda/conda-bld/linux-64/shiftx2-1.10-py27_0.tar.bz2

Skipping /opt/conda/conda-bld/linux-64/shiftx2-1.10-py27_0.tar.bz2
```

PR #816, opened 2017-09-26
[openpathsampling] 0.9.3

dwhswenson:
[OpenPathSampling 0.9.3](https://github.com/openpathsampling/openpathsampling/releases/tag/v0.9.3)


dwhswenson:
Long time since the last OPS update, so I'm not sure what all has changed here. But I'm having some trouble here with the Linux build. I'm getting the error:

```
/io/devtools/docker-build.sh: line 9: /opt/conda/bin/conda: No such file or directory
```

Is this a problem with the Docker image? It seems to happen early in the build, so I assume it's nothing that's specific to our recipe.

Also, getting the Mac build to work required that I install with `--single-version-externally-managed` -- is this the right way to do this? Otherwise, it looked like it was having trouble finding requirements during install (even though requirements had already installed, and all are listed in the `meta.yml`).

Finally, I was having trouble with AppVeyor yesterday, but I think that problems either on their end or with Continuum (kept getting HTTP errors when trying to download conda packages).

Lnaden:
Thats odd. Its almost like its not executing the correct set of actions or pulling the wrong docker image, which also does not make sense based on the travis log. I restarted the travis build to see if it was just a glitch. It could also be your fork has not been updated from the main repo in a while, which would cause issues if the last sync was before mid March.

Sorry for not getting around to this sooner

Lnaden:
Seems to have been a glitch, if you're done making changing, I'll merge it if you cannot

dwhswenson:
Yep -- one of those weird glitches where as soon as someone else looks at it, the computer decides to behave itself.

And yes, this should be ready to merge. Thanks!

PR #823, opened 2017-10-05
Add recipe for saltswap v0.5.1

bas-rustenburg:
This PR adds a conda recipe for the saltswap package by @gregoryross.

### DIFF
--- a/saltswap/bld.bat
+++ b/saltswap/bld.bat
@@ -1,0 +1,2 @@
+%PYTHON% setup.py install
+if errorlevel 1 exit 1
--- a/saltswap/build.sh
+++ b/saltswap/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+$PYTHON setup.py clean
+$PYTHON setup.py install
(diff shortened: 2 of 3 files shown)


bas-rustenburg:
Looks like for the time being, saltswap isn't windows compatible.

Lnaden:
Not quite, OpenMM is missing for Python 2.7 windows ( i think there was an issue for this), and is simply not built on 32 bit arch for windows. It does build for Py3k and 64bit.

bas-rustenburg:
Ah, good call. Do you know the syntax/have an example to restrict the builds to 64bit/py3?

Lnaden:
Half-ly?

```
build:
    skip: True # [py2k]
```
will skip the python 2 builds, but I'm not sure about the architecture, it may be something like:

```
build:
    skip: True # [(py2k and win) or win32]
```
Unfortunately, the conda docs are real light on valid skip strings, the best I could find for this is here:
https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#preprocessing-selectors

but I also found some other things here which may help, but is not quite what you need.
https://github.com/conda/conda-build/pull/1753

PR #827, opened 2017-10-12
Fix bug in saltswap meta.yaml

bas-rustenburg:
Accidentally left out closing ] has been added.


bas-rustenburg:
Looks to be building again. Merging.

PR #832, opened 2017-11-23
Update openforcefield for 0.0.3

davidlmobley:
Current version doesn't behave with modern networkx because we haven't done a new point release in months; I just created a new point release and this updates the recipe accordingly.

The repo has also moved.

### DIFF
--- a/openforcefield/meta.yaml
+++ b/openforcefield/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: openforcefield
-  version: 0.0.2
+  version: 0.0.3
 
 source:
-    git_url: https://github.com/open-forcefield-group/openforcefield.git
-    git_tag: 0.0.2
+    git_url: https://github.com/openforcefield/openforcefield.git
+    git_tag: 0.0.3
 
 build:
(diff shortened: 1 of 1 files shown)


davidlmobley:
@jchodera please review...

openforcefield/meta.yaml
```diff
@@ -20,7 +20,7 @@ requirements:
   run:
     - python
     - numpy
-    - networkx ==1.11
+    - networkx
```
jchodera: @davidlmobley : Can this release use *both* old and new `networkx` APIs, or can it only use the new `networkx` API?

davidlmobley:
Tests pass; anyone want to merge @Lnaden @jchodera ?

openforcefield/meta.yaml
davidlmobley: In principle it can use both, but we've had some issues with older versions still, so we should probably ask for version 2 and higher; that'd be `networkx >= 2`?

jchodera:
Thanks for the fix!

PR #833, opened 2017-11-27
Bump OpenMMTools to 0.13.4

Lnaden:


jchodera:
We're just running into travis timeouts with `osx`, so merging.

@lnaden: Do you have merge rights here?

Lnaden:
Yes, I do, I got sidetracked today

PR #835, opened 2017-11-28
Cache the Nvidia drivers on OSX to speed boot time

Lnaden:
This PR enables Travis' Caching feature to cache the CUDA Mac installers to speed up the OSX build time. The caching feature saves up to 20 min on the OSX builds.

I have also changed how the Homebrew/Cask update is handled to shave a few more minutes off the boot. I tried caching that as well, but ran into issues where it was not saving time.

Overall, this reduces the OSX boot time from ~40 min to 20 min.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ cache:
   directories:
     - $HOME/Library/Caches/Homebrew
     - $HOME/nvidia_cache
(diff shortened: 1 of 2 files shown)


Lnaden:
I should note that when we update the CUDA version, we will need to blow away the cache and download a new one. This can be done from the Travis control panel for the master branch

jchodera:
Awesome! Thanks!

PR #837, opened 2017-12-04
[openmm] OpenMM 7.2 beta build

jchodera:
@peastman : Take a look? I've merged in changes to this recipe from [the `dev` recipe](https://github.com/omnia-md/conda-dev-recipes/tree/master/openmm).


jchodera:
In particular: Should it be called `7.2.0` and appear in the `beta` channel? This way, we can test any codes that require the dependency `>=7.2`.

peastman:
> Should it be called 7.2.0 and appear in the beta channel?

Correct.

PR #838, opened 2017-12-04
Update omnia build system to CUDA 9.0 and force openmm 7.2 beta build

jchodera:
This PR makes some more changes in #837 
- [x] Update omnia build system to CUDA 9.0
- [x] Force openmm 7.2 beta to build and push (which may displace `dev`?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,8 @@ services:
 
 env:
   global:
-    - CUDA_VERSION: "8.0"
-    - CUDA_SHORT_VERSION: "80"
+    - CUDA_VERSION: "9.0"
+    - CUDA_SHORT_VERSION: "90"
   secure:
     - "K3LSjERPC+kKtF3qeE0i6r0LNvmliJcZIgFCS2jJv24KBYRLhRQit+WD0KgyuSOlEgW80zU1qbqxWWLRkVhUGtaZvKEnQHLS8ww1akYLBoWbEjJv4p2B+MiHo2fa7G8AYw5L28Ahmb4C6+/3/KjapX2K0xd5w7bSWVF1+iZPnts="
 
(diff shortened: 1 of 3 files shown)


jchodera:
We forgot to update the build test:
```
===== testing package: openmm-7.2.0-py27_0 =====

running run_test.py

  File "/Users/travis/anaconda/conda-bld/openmm_1512375518239/test_tmp/run_test.py", line 34, in <module>

    assert openmm.Platform.getOpenMMVersion() == '7.1.1', "openmm.Platform.getOpenMMVersion() = %s" % openmm.Platform.getOpenMMVersion()

AssertionError: openmm.Platform.getOpenMMVersion() = 7.2
```

Lnaden:
Odd. This appears to be trying to compile against all the different Numpy versions, but then not relabeling the builds, despite only needing *a* version of numpy installed, so it times out since it winds up building OpenMM 9 times on travis.

jchodera:
Any ideas?

jchodera:
We can merge it anyway and get the beta built and then debug from there.

Lnaden:
> We can merge it anyway and get the beta built and then debug from there.

Just to get something up, sure, probably a good idea. Might result in a few builds we have to delete before converting them to main.

> Any ideas?

Its probably something in conda or conda-build that changed which broke the way the conda-build-all script handles dependency processing. It might just be as simple as its returning a list of 9 blanks (hence why its not trying to upload `openmm_pyX.Y_npA.B.tar.gz` packages, which would be simple enough to fix, but annoying to debug. Just a guess though

jchodera:
Can you merge @lnaden? We'll watch what happens

Lnaden:
Odd, I can't replicate this multiple-builds for the same version error anymore, it also does not look like its in the travis builds anymore. So I'm going to chalk it up to some oddity with the conda-servers? We can reassess if it crops up again.

PR #839, opened 2017-12-04
[yank] release 0.19.3

andrrizzi:


andrrizzi:
It looks like I'm getting the same timeout error observed in #838 . I'd proceed with the release anyway if there are no objections.

jchodera:
I think I just have to retrigger the build a few more times until everything gets built.

PR #840, opened 2017-12-04
[openmm] Build OpenMM 7.2 beta as openmm-beta in main label; revert openmm recipe

jchodera:
This will push the OpenMM 7.2 beta to `openmm-beta`, like we did for previous betas.

This avoids the issue where pushing to `openmm` under the `beta` label overwrites the `dev` label and vice-versa.

Installation of the beta would require:
```bash
conda remove --yes openmm
conda install --yes openmm-beta
```

### DIFF
--- a/openmm-beta/bld.bat
+++ b/openmm-beta/bld.bat
@@ -1,0 +1,22 @@
+mkdir build
+cd build
+
+set FFTW=C:\Miniconda3\pkgs\fftw3f-3.3.4-vc14_2\Library
+set APPSDK=C:\Program Files (x86)\AMD APP SDK\2.9-1
+"C:\Program Files\CMake\bin\cmake.exe" .. -G "NMake Makefiles JOM" -DCMAKE_INSTALL_PREFIX=%PREFIX% -DCMAKE_BUILD_TYPE=Release -DOPENMM_GENERATE_API_DOCS=ON ^
+    -DOPENCL_INCLUDE_DIR="%APPSDK%\include" -DOPENCL_LIBRARY="%APPSDK%\lib\x86_64\OpenCL.lib" ^
+    -DFFTW_INCLUDES="%FFTW%/include" -DFFTW_LIBRARY="%FFTW%/lib/libfftw3f-3.lib"
+
+jom install
+jom PythonInstall
+jom C++ApiDocs
+jom PythonApiDocs
+REM jom sphinxpdf
+jom install
+
+mkdir openmm-docs
+robocopy %PREFIX%\docs openmm-docs * /e /move
+mkdir %PREFIX%\docs
+move openmm-docs %PREFIX%\docs\openmm
+mkdir %PREFIX%\share\openmm
+move %PREFIX%\examples %PREFIX%\share\openmm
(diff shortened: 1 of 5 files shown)


Lnaden:
If this is what we did last time, then it may be what we have to do. We could also bump the version on the `dev-recipes` to something higher such as 7.2.1, so there would not be a clash and would avoid the `openmm-beta` recipe which has to get built, then deleted later

jchodera:
> We could also bump the version on the dev-recipes to something higher such as 7.2.1, so there would not be a clash and would avoid the openmm-beta recipe which has to get built, then deleted later

Ooh, good idea. That would allow people to upgrade in-place without mucking about with deleting `openmm`, installing `openmm-beta`, and then reinstalling all the things that depended on `openmm` that were wiped out by the uninstall.

@peastman : Which would you prefer?

peastman:
If we really need two different recipes for dev and beta, is there some way we could at least avoid having so much duplication?  A lot of the files are identical, so could we contrive to have only one copy of them?

peastman:
Can you explain exactly what the two options are, and what the advantages/disadvantages of each are?

Lnaden:
Problem: We can't have a separate `dev` and `beta` version of openmm with the same package name because they overwrite each other. Any one package can have any number of labels, but the base package name is the starting point. 

Option A:
Do what we did last time. Create a new package on the main `conda-recipes` channel called `openmm-beta`. 
Pro: 
- Unique package which we can adjust and tinker with. 
- Its the solution which worked last time

Con: 
- Requires a different package name when installed through conda: `openmm-beta` vs `openmm.` 
- Requires us to delete the `openmm-beta` by hand when we are done with it. 
- Could cause confusion if you have `openmm` and `openmm-beta` installed at the same time.

Option B:
Bump the `conda-dev-recipes` version of `openmm` to a version higher than `7.2`
Pro:
- Easier to target specific package version while enforcing a single installable version on any given conda environment
    - `conda install -c conda-forge -c omnia openmm` <- main stable version
    - `... -c omnia/label/dev openmm` <- nightly dev version
    - `... -c omnia/label/beta openmm` <- 7.2 beta built by changes to the `openmm` recipe in the conda-recipes repo
- No new packages when we have to then delete

Con:
- Don't know what the appropriate bumped version number should be
- We'll have to delete the old dev version from the cloud

peastman:
How will each of these interact with other packages?  For example, as soon as we get the OpenMM beta up, I'll want to create a beta of PDBFixer that depends on the OpenMM beta.  And then I'll want to create a beta of OpenMM-Setup that depends on the betas of both OpenMM and PDBFixer.  And then users need to be able to type some command of the form `conda install (...) openmm-setup` and get the betas of all three packages (*not* the more recent dev versions).  Would either of these approaches present difficulties for that?

jchodera:
@peastman : OK, we should do what @Lnaden suggests and 
* update the `dev` builds to 7.2.1 `dev` label on `openmm`
* push the `beta` builds to 7.2.0 `beta` label on `openmm`

That way, all you have to do is make sure you've added the `beta` label (but NOT the `dev` label!) and have your OpenMM-Setup and PDBFixer packages require `>=7.2.0` if they need 7.2.0 specific features. 

I'll close this PR and bump the `dev` package.

PR #842, opened 2017-12-05
Update PDBFixer to 1.5

peastman:
@Lnaden could you confirm my changes are correct to generate the beta of 1.5?

### DIFF
--- a/pdbfixer/meta.yaml
+++ b/pdbfixer/meta.yaml
@@ -15,6 +15,7 @@ build:
 extra:
   upload: beta
   force_upload: False
+  include_omnia_label: beta
 
 requirements:
   build:
--- a/pdbfixer/meta.yaml
+++ b/pdbfixer/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: pdbfixer
-  version: 1.4
+  version: 1.5
 
 source:
-  fn: 1.4.tar.gz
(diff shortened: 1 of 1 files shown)


Lnaden:
That should upload PDBFixer 1.5 to the `beta` tag on the anaconda cloud.

Lnaden:
I don't think however that the conda-build-all script will actually find the openmm version because openmm 7.2 is on `beta`, and the build script does not look for the `beta` tag.

peastman:
What do I need to do to make it build the PDBFixer beta against the OpenMM beta?

Lnaden:
I don't think we've ever engineered a solution for that. I think this is the first time it has come up. It should be possible through to extend the conda-build-all script to look in the `extra` section for a flag we can engineer though

peastman:
That could work, though I have no idea how to do it.  Is that something you can do?  The issue here is that OpenMM, PDBFixer, and OpenMM-Setup all basically move as a unit.  Having created an OpenMM beta, we now need a PDBFixer beta that depends on it, and an OpenMM-Setup beta that depends on both of the others.

Lnaden:
I'm not 100% sure either since I did not make the original script, but I can take a stab at that tonight, might take me a couple hours to figure something out though

peastman:
Thanks!

Lnaden:
xref #843 which would enable the features discussed here

jchodera:
Thanks!

PR #843, opened 2017-12-05
Enable ability for packages in Omnia to build against a specific label.

Lnaden:
To any meta.yaml file, add the entry:
```yaml
extra:
   include_omnia_label: LABEL
```

where `LABEL` is replaced with any of the `STANDARD_LABELS` found in the `conda-build-all file`.
These tells the build script to prioritize searching in `-c omnia/label/LABEL` just behind `conda-forge`
so packages can build off of other omnia tags, such as beta.

This only searchs omnia tags to prvent someone from injecting nasty code from untrusted channels.

Enables #842 to proceed for OpenMM 7.2 beta testing

cc @jchodera @peastman
 Could you both take a look at this and see if this makes sense for to you?

### DIFF
--- a/conda-build-all
+++ b/conda-build-all
@@ -403,6 +403,17 @@ def build_package(m, python, numpy, args):
     cmd = [conda_build_cmd(), '-q', '--python', python, '--numpy', numpy, '--quiet']
     if args.notest:
         cmd.append('--no-test')
+    # Check if there is an omnia label to build against
+    if m.get_section('extra') and 'include_omnia_label' in m.get_section('extra'):
+        omnia_label = m.get_section('extra')['include_omnia_label']
+        if omnia_label not in STANDARD_LABELS:
+            print('Cannot build package {} against label "{}", must be one of the standard labels: {}.\n'
+                  'Falling back to using "main" only.'.format(m.path, omnia_label, STANDARD_LABELS))
+        else:
+            # Add conda forge to search AHEAD of omnia/label/LABEL otherwise the omnia channel will be higher
+            # priority than the conda-forge channel, which may grab very old build which we still have on the
+            # omnia channel. Its an edge case, but a critical one.
+            cmd.extend(['-c', 'conda-forge', '-c', 'omnia/label/{}'.format(omnia_label)])
     cmd.append(m.path)
     retcode = call(cmd)
     list_package_contents(m, python, numpy)


Lnaden:
This will only affect the specific package which has this specific entry, additional packages remain unaffected. Effectively severing as a single package, single tag use of the `--check-against` argument to the file as a whole

peastman:
Sounds like a good solution.

Lnaden:
Okay, if the tests here pass to make sure I did not break something else outside of my local testing, we can merge this in, then you'll just need to merge the upstream master into your PR to get the changes to take effect.

jchodera:
Looks like a travis error:
```
OSError: [Errno 28] No space left on device: '/var/folders/vy/rcv48w3j4w79llzf_x6qnvw40000gn/T/tmp_lql5y53'
```
I'll retrigger.

jchodera:
> That is a problem with the cache actually. But I'm not sure how to fix it.

The travis cache?

Lnaden:
Yes? But was that on this issue or a different one? I'm not seeing the error here

Lnaden:
and it should have only been on OSX

jchodera:
https://travis-ci.org/omnia-md/conda-recipes/jobs/312132147#L4123-L4145

Lnaden:
Yeah, that. Something with the cache causes the disk to run out of space sometimes, which is really annoying, but I have not had time to get around to Travis Support for help with this problem. We could disable the cache for now, but then the osx builds go back to being slow and requiring multiple re-triggers anyways to.

The problem is the tarballs for the Nvidia drivers are 1.2GB, and then we unzip them, but dont remove the tarballs before we build (so they can be cached at the end). We might be able to change it so only the untar'd data is cached, but I don't know where those files wind up, and where they need to go such that we can both cache exclusively them, and OpenMM can build against it. @peastman any ideas?

peastman:
No idea.  Where's the code that downloads and installs them?

Lnaden:
https://github.com/omnia-md/conda-recipes/blob/master/devtools/osx-build.sh#L29-L39

peastman:
It looks like it should be in whatever directory that script is run from.  $HOME maybe?  Or does it get run from inside the devtools directory?

peastman:
Here it is.  From the log of the job that failed:

```
+mkdir -p /Users/travis/nvidia_cache
+cd /Users/travis/nvidia_cache
+'[' -f cuda_mac_installer_tk.tar.gz ']'
+'[' -f cuda_mac_installer_drv.tar.gz ']'
+sudo tar -zxf cuda_mac_installer_tk.tar.gz -C /
+sudo tar -zxf cuda_mac_installer_drv.tar.gz -C /
```

Lnaden:
The linux side passed, so I am going to merge this and we can fix the cache problem separately

@peastman merge master into your fork's branch and then repush and #842 should start with these changes. You'll need to add the `include_omnia_label: beta` to the `extra` section of the meta.yaml file and it should allow PDBFixer to build against the OpenMMBeta

peastman:
Done, thanks!

PR #844, opened 2017-12-06
Update OpenMM-Setup to 1.1

peastman:


peastman:
Travis is working for Python 3.6 but failing for 2.7 and 3.5.  I don't understand why.

https://travis-ci.org/omnia-md/conda-recipes/jobs/312614706#L2085-L2105

Any idea what's going on?

jchodera:
Do you have a dependency called "itsdangerous"?

https://travis-ci.org/omnia-md/conda-recipes/jobs/312614706#L2228

peastman:
Not directly, though it's always possible one of the libraries it uses depends on that.

```
requirements:
  build:
    - python
    - setuptools
    - flask
    - openmm
    - pdbfixer >=1.5
```

Lnaden:
The package `flask` is pulling in `itsdangerous`, and then trying to install it again through `pip` for some reason, which then causes the builds to fail.

peastman:
It looks like it's some sort of cryptographic library, so it's probably required by Flask.  But it's been around for quite a while, and all indications are that it's supported on older Python versions.

https://pypi.python.org/pypi/itsdangerous/0.24

peastman:
Should I try listing it directly as a dependency in meta.yaml?

Lnaden:
In theory that should fix it, although I don't know why it isnt resolving, Maybe the minimum version in conda for flask needs a lower version than the pip version? that would cause this problem

peastman:
No, that didn't work.

Lnaden:
I'm looking at it now, I'll see if I can find where the dependency chain is broken

Lnaden:
Found it. Bad news, and good news.

The bad: Its the `itsdangerous` package through `flask` on conda-forge and specifically its `noarch` version is being fetched by all python versions, but not recognized by `setuptools` as a package installed called `itsdangerous`, so it then tries to pull it through `pip`. I don't know why, and neither does the person who found the exact same problem [on the itsdangerous feedstock](https://github.com/conda-forge/itsdangerous-feedstock/issues/9)

The Good: The person who reported this issue also engineered [a work around you can see here](https://github.com/jochym/ase-feedstock/commit/7c387863102751dfa498e70b3a610a508fdc0c49). the short of it is, in the `requirements: build` block of the meta.yaml, add the following:
```yaml
requirements:
  build:
    - itsdangerous 0.24 py27_1  # [py27]
    - itsdangerous 0.24 py35_1  # [py35]
    - werkzeug 0.12.2 py35_0  # [py35]
    - werkzeug 0.12.2 py27_0  # [py27]
```

And that works around the issue. you'll need to remove the pin to `itsdangerous` you added previously.

Lnaden:
I posted on `itsdangerous`'s feedstock, so hopefully it bumps the issue there.

peastman:
Perfect!  I just committed the change.  We'll see if it works.

peastman:
All tests passed.

peastman:
According to https://anaconda.org/omnia/openmm-setup/files, on Windows it seems to have only built a Python 3.5 version, not a 3.6 version.  Any idea why?

Lnaden:
` skip: True  # [win32 or (win64 and not py35)]` in the meta.yaml file

peastman:
Ah, right.  Oops.  Let me go fix that.

PR #847, opened 2017-12-15
Fix to building OpenMM on Windows

peastman:


Lnaden:
You'll have to remove the [`skip: True # [win]`](https://github.com/peastman/conda-recipes/blob/1e7cae27a7ac721afda843c295f45e6f530b29bf/openmm/meta.yaml#L12) line to get the windows builds to fire. I think because there would be nothing left to skip, you'll need to either comment out or remove the whole line.

We have Windows build against 32- and 64-bit architecture though, so you may need to just change the skip to pass over 32 bit unless we actually want a 32-bit OpenMM to run

peastman:
Right now I just change that manually when doing the builds.  See https://github.com/pandegroup/openmm/wiki/Building-conda-packages-on-Windows.  I was just trying to simplify one other step in the process.

I'm not actually sure why we have it skipping Windows builds.  I assume there's some difficulty with doing them on AppVeyor?

Lnaden:
> I was just trying to simplify one other step in the process.

Ah, that makes sense.

> I'm not actually sure why we have it skipping Windows builds.

As I recall, there was some problem with needing a specific Visual Studio version, but I would need to dig around the issues last time OpenMM did a release

peastman:
Is it alright to merge this?  Or will that cause it to do a new build and overwrite the existing betas?

Lnaden:
It should not cause a rebuild because `force_upload` is `False` and the version has not changed

jchodera:
Sorry, hadn't realized it was waiting on us!

peastman:
Thanks!  I could have merged it myself, but I wanted to make sure that wouldn't have any unintended consequences.

PR #852, opened 2018-01-11
Fixed error building OpenMM on Windows

peastman:
Because *of course* using Windows style paths on Windows doesn't work correctly...

### DIFF
--- a/openmm/bld.bat
+++ b/openmm/bld.bat
@@ -3,7 +3,7 @@ cd build
 
 set FFTW=C:\Miniconda3\pkgs\fftw3f-3.3.4-vc14_2\Library
 set APPSDK=C:\Program Files (x86)\AMD APP SDK\2.9-1
-set CUDASDK=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v9.0
+set CUDASDK=C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v9.0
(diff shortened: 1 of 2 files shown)


peastman:
Updating the recipe hasn't caused new builds to be generated.  https://anaconda.org/omnia/openmm/files.  Any idea why not?  @jchodera @Lnaden

Lnaden:
Yes, and it was a problem only affecting the dev channel which I have been trying to work with the Travis people to fix, but now its affecting this one too. Fun

Lnaden:
From their replies, this is what we are currently at:

"something, somewhere in the relevant process is switching stdout to be non-blocking (see comments in https://github.com/travis-ci/travis-ci/issues/4704)"

Do you know of anything changed since about Dec 12th which would have made stdout be non-blocking?

peastman:
Do you mean in this repository or the OpenMM repository?  The changes here are trivial and shouldn't have any effect.  There have also only been a handful of changes to OpenMM since then and I don't see a reason they would make a difference: https://github.com/pandegroup/openmm/commits/master.

Lnaden:
Probably OpenMM since that is the build which throws the error. The fix I have seen people try to do from that issue I linked was to add this line to the head of their `.travis.yml` script:

`- python -c 'import os,sys,fcntl; flags = fcntl.fcntl(sys.stdout, fcntl.F_GETFL); fcntl.fcntl(sys.stdout, fcntl.F_SETFL, flags&~os.O_NONBLOCK);'`

So its something we could try, even if I don't understand why that works

jchodera:
Can you take a stab at that @lnaden?

PR #853, opened 2018-01-15
Experimental: Try setting the non-blocking flag in Python

Lnaden:
May allow large sdtout flood not to cause travis to crash

xref #852


Lnaden:
This did not work, will keep digging

PR #854, opened 2018-01-16
Use old travis trusty image since new one breaks our builds.

jchodera:
See: https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch

This hopefully fixes the issue noted by @peastman in https://github.com/omnia-md/conda-recipes/pull/852#issuecomment-357307310

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+# Use old trusty image since new one breaks our builds
+# See: https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch
+group: deprecated-2017Q4
+
 os:
   - linux
   - osx


Lnaden:
My guess is it won't, see https://github.com/omnia-md/conda-dev-recipes/pull/111 where it didn't

Lnaden:
Also, just got an email from the Travis support team, posting verbatim since I don't have time to debug.

----

Thanks for the update and I'm sorry for all this back and forth.

We've been investigating into this further and we noticed that scenario that you're experiencing is similar to other reports we've received lately and were caused by a regression in one internal tool that process logs to make sure secrets are not accidentally exposed there. I'm sorry for the inconvenience this has caused you and your team.

To sort this out we changed some configuration on our end, and we believe that this might get your builds working again, without getting this abrupt interruption of the log anymore.

Could you please try to restart any affected build (or start a new one) and let us know how that works out?

Thanks in advance and let's follow-up upon your reply.

jchodera:
Restarting the `master` build!

jchodera:
Oh, I see you've already restarted them!

It does seem to be *building* openmm, but not uploading the new packages.

Lnaden:
I actually had not restarted them, the link I sent was from the `dev` package and using the old image here may not have been required. I'm also not a fan of relying on the old image if we can avoid it

jchodera:
Agreed, but the reversion managed to get the packages posted. We can now safely look into how to adapt to the new image without holding up OpenMM testing.

peastman:
Thanks!

peastman:
But it looks like it didn't build a Python 3.6 version for macOS?

Lnaden:
Global build time was reached after py27 and py35, I'l re-trigger and it should build 3.6

peastman:
Still no sign of it.

jchodera:
Looks like the `osx` builds are backlogged: https://travis-ci.org/omnia-md/conda-recipes

Lnaden:
Yeah, looks like Travis is having some problems today, status says they have having linux and OSX issues, causing a backlog of the whole system it looks.

https://www.traviscistatus.com/

jchodera:
I just restarted the `osx` build because it died without running last night.

PR #857, opened 2018-01-29
Added forcebalance

leeping:


forcebalance/meta.yaml
```diff
@@ -0,0 +1,53 @@
+package:
+  name: forcebalance-dev
```
Lnaden: Do you actually want to call this `forcebalance-dev`? That is how people will locate the package, they will have to call `conda install forcebalance-dev` to access it. 

If you want to upload this to the `dev` label, we have another way to do it.

forcebalance/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  version: "1.5.1"
+
+source:
+  url: https://github.com/leeping/forcebalance/archive/v1.5.1.tar.gz
+  fn: v1.5.1.tar.gz
+
+build:
+  number: 0
+  skip: True # [py2k]
```
Lnaden: I know FB recently became Py3 compatible, but do you truly want to skip the Python 2.7 builds?

leeping: Oops - I didn't know that's what it meant!

forcebalance/meta.yaml
leeping: I'll remove the -dev label now.

forcebalance/meta.yaml
leeping: Should this be changed to `skip: False`?

forcebalance/meta.yaml
Lnaden: We can have it upload to the `dev` label pretty easily actually, if you don't want it on the `main` channel

forcebalance/meta.yaml
Lnaden: You can just remove this line if you dont want any builds skipped. Its optional

PR #858, opened 2018-01-30
Build OpenMM 7.2rc2

peastman:
@jchodera @Lnaden is this correct, or do I also have to set `force_upload` to `True`?

### DIFF
--- a/openmm/meta.yaml
+++ b/openmm/meta.yaml
@@ -4,11 +4,11 @@ package:
 
 source:
   git_url: https://github.com/pandegroup/openmm.git
-  git_tag: 7164109
+  git_tag: f481bd0
 
 
 build:
-  number: 1
+  number: 2
   skip: True # [win]
 
 extra:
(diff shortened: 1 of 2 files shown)


jchodera:
Incrementing the build number *should* force re-upload.

Lnaden:
Yes, the build numbers let you rebuild when there have been changes to the code base, but not the version. If someone tries to install "the most recent version," or even the same version number, it will prioritize the higher build number

peastman:
Thanks!

One of the travis builds failed right at the start:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

I'll try restarting it.

jchodera:
Whoops, that gets me every time!

peastman:
It uploaded two of the Mac installers fairly soon after I merged this, but there's been nothing else for a few hours.  How do I check what it's doing?

https://anaconda.org/omnia/openmm/files

jchodera:
Check the Travis output for the master branch:
https://travis-ci.org/omnia-md/conda-recipes

Lnaden:
It looks like it hit the maximum allowed Travis wall clock time midway through the Python 3.6 builds, so those versions never finished and never pushed. If we were to restart the jobs, it should correctly see that Python 2.7 and 3.5 are built, skip those, and proceed to the python 3.6 build as desired.

peastman:
I restarted them.

peastman:
Everything built.  Thanks!

PR #861, opened 2018-02-25
[openpathsampling] 0.9.4

dwhswenson:
[OpenPathSampling 0.9.4](https://github.com/openpathsampling/openpathsampling/releases/tag/v0.9.4)

### DIFF
--- a/openpathsampling/meta.yaml
+++ b/openpathsampling/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: openpathsampling
-  version: 0.9.3
+  version: 0.9.4
 
 source:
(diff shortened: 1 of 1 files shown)


dwhswenson:
@Lnaden : looks like everyone else is merging their own here; I don't have access, and you merged the last OPS update -- mind doing this one, too?

(Near future, OPS will move OpenMM to an optional dependency and we'll move to conda-forge -- in other words, before you offer, I don't think I need commit access here).

Lnaden:
Got it! And thanks for also answering my follow up question!

PR #864, opened 2018-03-07
Update packmol to 18.013

Lnaden:
Also drops the libgfortran 1.0 pin

### DIFF
--- a/packmol/meta.yaml
+++ b/packmol/meta.yaml
@@ -1,15 +1,15 @@
 package:
   name: packmol
-  version: "1!17.221"
+  version: "1!18.013"
 
 source:
   fn: packmol.tar.gz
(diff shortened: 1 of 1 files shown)


mcs07:
Can the libgfortran 1.* pin be removed from openmoltools and solvationtoolkit also?

Lnaden:
We did for the main OpenMolTools repo, so we can cut another release for that. Solvationtoolkit also removed it a while ago, but needs a 0.4.3 release it looks like. @davidlmobley could we get a 0.4.3 release of Solvation Toolkit even if just to remove the libgfortran 1.0 dependency?

davidlmobley:
Yes, should be able to do that. Thanks for the heads-up, putting it on the list.

PR #865, opened 2018-03-08
Update OpenMolTools to 0.8.3

Lnaden:
This is the last Python 2.7 release for this package


Lnaden:
I just realized I never actually merged this...

PR #866, opened 2018-03-22
Update Parmed version to 3.0.1

nividic:


nividic:
@jchodera @Lnaden I'm trying to update the Parmed version to the latest release but I'm having some issues. Can someone help me please?

Lnaden:
It looks like there may be a problem with reaching the conda servers. I am going to try to re-fire the builds and see if it was just a fluke.

nividic:
@Lnaden Thank you so much

Lnaden:
Okay,  I know  what's wrong. 

Quick fix: @peastman is it okay if we remove the `rc` tag from the OpenMM 7.2.1 release on the anaconda cloud now?

Long (and correct fix): #867

peastman:
> Quick fix: @peastman is it okay if we remove the rc tag from the OpenMM 7.2.1 release on the anaconda cloud now?

If you do that, you should probably remove *all* rc tags.  Otherwise if someone comes across the forum post and types the command to install the release candidate, they'll get a release candidate for the previous version instead, which certainly isn't what we want.

But what will the consequences of that be?  We'll be left with some builds that have no label attached to them at all.  How will people identify what those builds were (release candidates)?

Lnaden:
> they'll get a release candidate for the previous version instead, which certainly isn't what we want.

> But what will the consequences of that be? We'll be left with some builds that have no label attached to them at all. How will people identify what those builds were (release candidates)?

Both of those are very good points I had not thought of in a thought for a quick solution. Scratch that request. 
I did just push the actual fix to the `conda-build-all` file in #868, so this should not be a problem anymore.

@nividic Please try merging in the current master to your fork and re-uploading. That should fix the build issues this is having.

nividic:
@Lnaden I was checking the Parmed conda packages on omnia...I can see just the windows versions but not the linux and osx ones https://anaconda.org/omnia/parmed/files. Am I looking in the wrong place?

nividic:
@Lnaden they are coming out right now!

Lnaden:
The files sometimes take a bit to publish since the CI re-runs on the master branch to do the building and uploading. The OSX builds especially can be slow as those are limited by the number of globally concurrent Travis OSX instances that are running.

Do let me know if some builds don't show up and I'll look into it!

PR #873, opened 2018-04-10
Update protons to 0.0.1a2

bas-rustenburg:
This updates protons to the latest tagged version.


bas-rustenburg:
Not sure how that broke the builds. Perhaps something changed on the conda side of things..

PR #875, opened 2018-04-11
 Pin conda to a max to not break its API

Lnaden:
The actual fix is something I am working on so we have conda-build 3 compatibility and we can stop pinning to conda and conda-build, but many, many, many things changed and it will take a while to fix.

Since I don't know how long that will take, this is a stopgap to pin the conda version to stop this cycle of band-aid'ing while I get a proper fix in.


Lnaden:
I think this should fix the linux and OSX builds. Windows is still having some problems I will need to fix at a later time when I can see more what downgrading from conda 4.4+ to 4.3.X does to the path.

PR #877, opened 2018-04-21
 [fahmunge] 0.3.0

jchodera:
Update `fahmunge` for 0.3.0 release.


jchodera:
Something is wrong here. Fixing issues with the `fahmunge` 0.3.0 release first.

PR #879, opened 2018-05-03
Update Lunrsearch

Lnaden:
Thanks @rmcgibbo 

cc pandegroup/openmm#2048


Lnaden:
Failing is due to other windows problems I need to resolve at some point. otherwise passes

PR #880, opened 2018-05-10
OpenMM 7.2.2

peastman:


peastman:
All the AppVeyor builds are failing:

```
conda config --add channels omnia
'conda' is not recognized as an internal or external command,
operable program or batch file.
```

Is this a known issue?

Lnaden:
Yes, Its been on my to-do list but has been low priority. Its been a problem since I had to pin to an eariler version of `conda` because we cant update `conda-build` yet, and I need to fix the `conda-build-all` script. I can make it a priority if need be

peastman:
That's ok.  I just wanted to make sure it wasn't something I should worry about.

PR #890, opened 2018-06-23
add skorch

cxhernandez:
Adding skorch library and patching it to be used with osprey

### DIFF
--- a/skorch/meta.yaml
+++ b/skorch/meta.yaml
@@ -1,0 +1,44 @@
+package:
+  name: skorch
+  version: 0.2.0
+
+source:
+  git_url: https://github.com/dnouri/skorch.git 
+  git_tag: v0.2.0
+  patches:
+    - 0001-Compatibility.patch
+
+build:
+  skip: True  # [win or osx or py27]
+
+extra:
+  maintainers:
+   - cxhernandez
+
+requirements:
+  build:
+    - python
+    - scipy
+    - numpy
+    - scikit-learn
+    - tabulate
+    - tqdm
+
+  run:
+    - python
+    - scipy
+    - numpy
+    - pytorch
+    - scikit-learn
+    - pyyaml
+    - tabulate
+    - tqdm
+
+test:
+  # Modify tests to get around bug (remove in future versions)
+  #imports:
+  #  - skorch
+
+about:
+    home: https://github.com/dnouri/skorch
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -57,3 +57,8 @@ script:
         export NVIDIA_CACHE=$HOME/nvidia_cache;
         bash devtools/osx-build.sh;
     fi
+
+addons:
+    ssh_known_hosts:
```
cxhernandez: This doesn't seem to be helping with the ssh issues

skorch/meta.yaml
```diff
@@ -0,0 +1,38 @@
+package:
+  name: skorch
+  version: 0.3.0dev
+
+source:
+  git_url: https://github.com:dnouri/skorch.git 
+  git_tag: 00f53ecf9455df702c7fe686e238ac7641567e03
```
jchodera: If you really need this particular git hash instead of a release, follow [this example from conda-dev-recipes](https://github.com/omnia-md/conda-dev-recipes/blob/master/openmm/meta.yaml#L6-L7), or even better, put this recipe there so your dev build will be built nightly!

skorch/meta.yaml
```diff
@@ -0,0 +1,38 @@
+package:
+  name: skorch
+  version: 0.3.0dev
```
jchodera: Since this isn't a release you might want to include the git shortened hash in the version number?

jchodera:
I can't quite tell from the error message, but it looks like there may be some dependencies in [`setup.py`](https://github.com/dnouri/skorch/blob/master/setup.py#L10-L27) that aren't in your [`meta.yaml`](https://github.com/cxhernandez/conda-recipes/blob/74b522375fd1acbdfb265553161ee9be2df3641c/skorch/meta.yaml#L16-L29). Is [`tabulate`](https://anaconda.org/conda-forge/tabulate) something you should add?

cxhernandez:
Ah, yeah must be that, now that I'm looking at `setup.py`. They didn't list that one in their `requirements.txt` 😞

jchodera:
Looks like we're missing tqdm somehow:
```
RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml  url=https://pypi.org/simple/tqdm/
```

PR #907, opened 2018-08-10
Build on Python 3.7

Lnaden:
I know this will not work for all packages, but we need to start supporting it as Python 3.7 has been officially released


Lnaden:
I am just going to merge this now, so we can start getting builds. We can always reverse it after the fact later if its a problem, but the "tests" here will never pass until merged because everything will time out.

PR #911, opened 2018-08-28
Updates for OpenMM 7.3 beta

peastman:


jchodera:
@peastman : We're building the OpenMM 7.3 beta in [`conda-dev-recipes`](https://github.com/omnia-md/conda-dev-recipes), not this repo, since we haven't set up the infrastructure to build with multiple CUDA images yet here.

jchodera:
Though I realize I need to prevent the force-updating of those packages in `conda-dev-recipes`. Right now, it looks like it's rebuilding the beta nightly, which we *don't* want.

peastman:
We still need to keep this repository up to date though.  Especially because the Windows builds *are* based on this repo.

jchodera:
Agreed! Thanks! Will merge, then.

jchodera:
Whoops: If I merge this, it will rebuild CUDA for `osx` and `linux` with CUDA 9.0 and overwrite the `beta`. Let me update the build infrastructure first before we merge this.

jchodera:
I'll tackle #912 and then merge this in.

jchodera:
We'll also have to be sure to check against the `beta` channel first so that this doesn't retrigger beta builds that are already on anaconda cloud.

peastman:
Any update?  It would be very helpful if this could be merged before we build the 7.3 release candidate.

jchodera:
Ugh, everything is broken again. Sorry for the delay here---I have to clean up the current mess first, but hopefully can get this checked in soon.

peastman:
Can this be merged?

jchodera:
@peastman: I've been fighting with the build systems over at http://github.com/omnia-md/conda-dev-recipes to build Python 3.7 since conda-forge has already deprecated Python 3.5 (which means, in effect, so have we, though I think we can still manage the 3.5 build). 

In order to build Python 3.7, we'll need to build the omnia linux anvil images. We can either wait for @Lnaden to finish https://github.com/omnia-md/omnia-linux-anvil/pull/12 or we can just tweak the existing ones to add the missing TeXLive packages needed for the latest sphinx to work.

Lnaden:
I won't be able to work on the anvil changes I have for a few days, so I would tweak the existing anvil to add the latex packages. I think you could just call `tlmgr` from the `build.sh` script for the purposes of speed, but I don't know if that needs sudo access and the anvil entry point does not have that. I think this was a thing at one point, but I don't remember.

Lnaden:
I also will want to do testing on the `dev` side for a while first to make sure the updated build script I have works before merging in the changes to the main recipe channel

jchodera:
> I think you could just call tlmgr from the build.sh script for the purposes of speed, but I don't know if that needs sudo access and the anvil entry point does not have that. I think this was a thing at one point, but I don't remember.

I've tried that without success.

peastman:
Can this be merged?  I need to build the release candidate, and it depends on these changes.

PR #919, opened 2018-11-01
[WIP] Upgrade build system to support py37 and multi-CUDA builds

jchodera:
I'm a bit worried that the multi-CUDA builds will be a disaster for `conda-recipes`, so we may need to restrict the building of packages that require multi-CUDA builds to some other repo.

This will test how bad things might be.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ os:
   - linux
   - osx
 
+
 # Supported osx/xcode versions: https://docs.travis-ci.com/user/languages/objective-c/#Supported-Xcode-versions
 # See also: https://blog.travis-ci.com/2016-10-04-osx-73-default-image-live/
 osx_image: xcode6.4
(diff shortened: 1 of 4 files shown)


jchodera:
The other alternative is to simply create `omnia-md/cuda-conda-recipes` and build OpenMM production versions (and other plugins that require CUDA) there.

Lnaden:
The variant system in Conda Build 3.0 can help with de-duplication, and if we enable travis to build on every Python Version, i don't think it will be too bad. See my thoughts on this in https://github.com/omnia-md/conda-dev-recipes/pull/154#issuecomment-428990519

jchodera:
@lnaden: Thanks! We need to do something to build OpenMM 7.3 in the near term, but this sounds great for long term. 

I think we will just use conda-dev-recipes to build the 7.3 rc, and update conda-recipes to add Python 3.7 builds for other packages, then move to your conda build 3.0 proposal.

jchodera:
I'm going to deprecate this and just add Python 3.7 support separately once OpenMM 7.3 is released (which supports Python 3.7).

PR #920, opened 2018-11-05
[openpathsampling] 0.9.6

dwhswenson:
Update for the most recent release (probably the last before 1.0!)

### DIFF
--- a/openpathsampling/README.md
+++ b/openpathsampling/README.md
@@ -1,5 +1,4 @@
-This is a recipe for building the current development package into a conda
-binary.
+This is a recipe for building a release into a conda binary.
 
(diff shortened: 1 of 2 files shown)


dwhswenson:
I think that the failures here are the same as on all builds since #907, and not a problem with our recipe. Could someone confirm? (and if that's the case, merge?) Thanks!

dwhswenson:
It looks like the OPS build here is going fine, except for Python 3.7 (which looks like the same problem openmmtools had in its most recent release.)

@jchodera : Could you look this over and merge? Thanks!

jchodera:
Thanks! We'll fix py37 once OpenMM 7.3 is released in a few days.

PR #922, opened 2018-11-28
Cleanup for python 3.7

jchodera:
This PR:
* removes packages that have been migrated to conda-forge
* adds numpy >=1.14 dependency (latest version provided for conda-forge python 3.7 builds) to those packages that need it to enable builds with python 3.7
* fixes some other typos and failures

I'm only partway through a cleanup.


jchodera:
Oof---the version of conda-build we're using doesn't know `py37`, we we can't skip `py37` builds for recipes that don't support them:
```
Warning: Treating unknown selector 'py37' as if it was False.
```
@Lnaden : Any chance you can help us finish the upgrade of our build infrastructure after we get this cleanup PR merged?

jchodera:
@cxhernandez : Do you still need the recipe for building new versions of `hmmlearn` in here? If so, could you add a PR to fix it for py37 support? Otherwise, I'll likely have to remove the recipe (not the existing packages).

jchodera:
Actually, `hmmlearn` has been migrated to conda-forge, so I can remove that too:
https://github.com/conda-forge/hmmlearn-feedstock

PR #935, opened 2019-02-05
[cerberus] 1.1

jchodera:


jchodera:
I've built the `cerberus` package manually with conda-build 3 and pushed it to `omnia`.

PR #939, opened 2019-03-16
[assaytools] Skip py37

jchodera:


jchodera:
@sonyahanson: It looks like skipping py37 won't work until we upgrade to conda-build 3.

Alternatively, we can migrate the recipe to `conda-forge`. Since this package doesn't rely on `openmm`, that may be the best idea. Would you be up for that?

sonyahanson:
Works for me!

jchodera:
OK, I can tackle this. 

We should cut a new release first, though! I think we're getting closer to "stable" code, and there have been 69 commits since the last release!
https://github.com/choderalab/assaytools/releases

PR #943, opened 2019-04-13
[parmed] 3.2.0

jchodera:
@swails : Can you check if this looks right to you? No changes in dependencies, right?

cc: https://github.com/ParmEd/ParmEd/tree/3.2.0


jchodera:
Everyone **Watch**in the repo gets an email!

PR #948, opened 2019-05-18
Update osx builds: Use xcode 9.4, target osx 10.10, skip CUDA/additional tex package installs

jchodera:
* Now that `xcode 6.4` has been deprecated:
image
I've updated builds to use [`xcode` 9.4](https://docs.travis-ci.com/user/reference/osx/#macos-version) and the target osx version to 10.10.
* We now install brew packages via the [`addons`](https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-macos) mechanism
* Since CUDA-dependent builds have moved to https://github.com/omnia-md/conda-recipes-cuda, I've commented out the CUDA toolkit installation (and additional TeX package install) for `osx` since this was causing problems with the docker image.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,8 @@
+
+# Use old trusty image since new one breaks our builds
+# See: https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch
+#group: deprecated-2017Q4
+
 os:
   - linux
   - osx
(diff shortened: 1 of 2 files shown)


jchodera:
We may need to switch to [this scheme](https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-macos) for installing `osx` packages via `brew`.

PR #965, opened 2019-08-05
Fix travis failures

jchodera:
* [assaytools] skip py37 since pymc2 not available for py37
* [protons] skip, since broken


jchodera:
It looks like skipping py37 won't work until we upgrade to conda-build 3.

PR #977, opened 2019-09-04
PDBFixer 1.6

peastman:


peastman:
The tests are failing for reasons that I think are unrelated to this change:

```
Can't build /io/assaytools due to environment creation error:
The following specifications were found to be in conflict:
  - assaytools 0.3.0 py37_1
```

Should I go ahead and merge this?

j-wags:
@peastman Assaytools always fails -- It won't affect your build. 

It looks like my openforcefield_forcefields package is also rebuilding for some reason. I probably need to debug that. That may slow down the builds, but shouldn't affect anything other than the speed.

This looks like it's good to merge.

peastman:
Thanks!

jchodera:
> It looks like my openforcefield_forcefields package is also rebuilding for some reason.

@j-wags : This is probably due to `conda-build-all` not checking against the package channel that `openforcefield_forcefields` was pushed to---you called it the `dev` channel, which is not checked by default:
https://anaconda.org/omnia/openforcefield_forcefields/files

I'd suggest either (1) using a channel that is checked by default, such as `rc` or `beta` (I think), or (2) if you need nightly dev builds, using the [`conda-dev-recipes`](https://github.com/omnia-md/conda-dev-recipes) repo.

PR #986, opened 2019-10-09
[propertyestimator] 0.0.5 (label: rc)

SimonBoothroyd:


SimonBoothroyd:
cc @j-wags

SimonBoothroyd:
> Hooray for more permissive dependency versions!

We'll see how this goes. `dask-jobqueue` has finished being rewritten so that most of the core aspects which had previously been prone to get out of sync with `distributed` have been pushed upstream so I'm hoping this won't be an issue any more.

> Would it also make sense to add the openforcefields repo as a dependency?

I don't think so. We currently only use the smirnoff99frosst repo for testing purposes, and now that the framework supports multiple different force fields outside of OFF I think it makes sense for it to be left as an optional thing.

j-wags:
Understood -- That makes sense. 

Travis builds look good. This is ready to merge. I'll let you click when ready.

PR #988, opened 2019-10-15
[openforcefield] 0.5.1 build 1

j-wags:
Adds requests and notebook as dependencies.

@jchodera Any objections? This will make the install slightly larger, but users currently have to install these separately to run all examples.

### DIFF
--- a/openforcefield/meta.yaml
+++ b/openforcefield/meta.yaml
@@ -32,7 +32,7 @@ requirements:
     - smirnoff99frosst
     - openforcefields
     - openmm
-    - networkx
+    - networkx==2.3
     - parmed
     - rdkit
     - ambermini
@@ -53,7 +53,7 @@ requirements:
     - smirnoff99frosst
     - openforcefields
     - openmm
-    - networkx
+    - networkx==2.3
     - parmed
     - rdkit
     - ambermini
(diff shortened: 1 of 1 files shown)


jaimergp:
If dependencies/installation size are an issue, one thing to consider would be having a separate `openforcefield-examples` repo like [yank is doing](https://github.com/choderalab/yank-examples).

jchodera:
Do we _install_ the examples now? 

OpenMM copies its examples to them to `$PREFIX/share/openmm/examples`, so we might want to copy our examples to `$PREFIX/share/openforcefield/`.

j-wags:
We do not copy the examples out right now, though I agree that would be a good idea. I don't love the idea of people poking around inside `~/miniconda3/envs/openforcefield/share/openforcefield`, but saying "install the toolkit once, then git clone the repo again" is pretty ridiculous, so I'll add those to the package.

I'm not sure about fully separating the examples from the toolkit at this time. Keeping them together (and in the same CI builds) right now ensures that we don't leave them in a broken state during a release, and the example files/dependencies aren't _too_ big yet.

j-wags:
Opened https://github.com/openforcefield/openforcefield/issues/437 per John's comment above.

jaimergp:
Multi-output recipes like [this one](https://github.com/conda-forge/ucx-split-feedstock/blob/master/recipe/meta.yaml#L26) could be a solution. They will be maintained in the same repo, but the conda recipe will produce two packages.

PR #989, opened 2019-10-30
[propertyestimator] 0.0.6 (label: main)

SimonBoothroyd:
cc @j-wags


SimonBoothroyd:
@j-wags thanks for re-triggering - things look like they've passed now.

PR #999, opened 2019-11-15
test whether py38 builds are possible

j-wags:


jchodera:
One issue here is that we're pinning `conda-build` to make our current infrastructure work:
> WARNING: conda-build appears to be out of date. You have version 2.1.17 but the
latest version is 3.18.11.

I have no idea if that old `conda-build` could handle python 3.8 builds...

jchodera:
It looks like some of these packages are actually building, but we also probably need to build OpenMM for python 3.8 first. That's over here: https://github.com/omnia-md/conda-recipes-cuda

PR #1010, opened 2020-01-27
Conda Build v3 revamp

jaimergp:
Following updates in `conda-dev-recipes`, I will start adapting (very conservatively) `conda-recipes` to the new CB3 infrastructure running on Azure.

# To do
- [ ] Azure gets triggered
- [ ] All recipes are compatible with the new infrastructure (no `extra` fields are needed)
- [ ] Uploads work correctly
- [ ] Decide whether to build `openmm` or not

# Changelog

* 2020.01.27 - Copied scripts from `conda-dev-recipes` (with uploads disabled), deleted Travis and Appveyor configs.


jaimergp:
Added a summary of the build status of the different recipes. 

I think we should also take some time to decide which packages are going to "make it" to the CB3 revamp (or optimistically to `conda-forge`). I guess that those who won't make it it's because:

* they were dependencies that are now available in CF. 
* they are no longer being maintained / developed.

Can we choose which ones to keep and which ones to deprecate? @jchodera @Lnaden

jchodera:
My understanding is that we have already removed recipes for packages available on conda-forge, and that remaining ones likely need to be migrated. Some may no longer be actively maintained, however.

What's the best way to mark what needs to transition and what should be deprecated? Should we mark this in the yaml files or in this PR?

Also, should we add a list of maintainers to the yaml file, or will it be understood the recipes left to maintain are up to us?

jaimergp:
I have moved those already in CF to `deprecated`. Then added a `CONDA_FORGE_VALID` blank file in those remaining under `recipes` if they don't seem to have an incompatible license.

jaimergp:
I'd like to include `extra.maintainers` too, wishing somebody else will volunteer to participate :) But yes, at first I see us as the only maintainers.

Issue #1006, opened 2020-01-12
Standardize on AmberTools instead of AmberMini


jchodera:
We should remove `ambermini` from the dependencies of all our projects, most critically:
* `openforcefield`
* `openmoltools`
and instead use `ambertools` from the Amber MD channel or `conda-forge`.

jchodera:
Thanks to @jaimergp's fantastic work, we now have ambertools built for both osx and linux on [conda-forge](https://anaconda.org/conda-forge/ambertools/files)!

We should move ahead with these changes ASAP to resolve these issues now.

jchodera:
Fixed by https://github.com/omnia-md/conda-recipes/pull/1009

PR #1011, opened 2020-01-28
[openmmtools] use conda-forge ambertools to provide parmed

jchodera:
This removes another instance of `parmed` as a dependency, since it is now in `conda-forge` ambertools.

cc #1006

### DIFF
--- a/openmmtools/meta.yaml
+++ b/openmmtools/meta.yaml
@@ -7,7 +7,7 @@ source:
     url: https://github.com/choderalab/openmmtools/archive/0.19.0.tar.gz
 
 build:
-  number: 0
+  number: 1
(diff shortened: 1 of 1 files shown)


andrrizzi:
I thought `parmed` is on `conda-forge` as well. And I don't think openmmtools depends on amber (`ambertools` is quite a big package). Is the problem that `ambertools` now supplies its own version of parmed which is in conflict with the one on conda forge?

jchodera:
Yes, conda-forge ambertools supplies its own version of parmed right now.

Eventually, we should have conda-forge ambertools *depend on* the conda-forge parmed, but this will take more work (with help from @jaimergp) to make work.

PR #1015, opened 2020-02-12
[openmoltools] Upgrade to 0.8.5

jaimergp:
@hannahbrucemacdonald

### DIFF
--- a/openmoltools/meta.yaml
+++ b/openmoltools/meta.yaml
@@ -4,7 +4,7 @@ package:
 
 source:
   git_url: https://github.com/choderalab/openmoltools.git
-  git_tag: 0.8.5
+  git_tag: v0.8.5
 
 build:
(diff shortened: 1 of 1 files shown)


hannahbrucemacdonald:
I don't know what the `number: X` does but LGTM

jaimergp:
That's the _build_ number. For example in `openmm` dev:

```
win-64/openmm-7.5.0-py38_cuda102_0.tar.bz2
                                 ^
```

It's used to indicate that we have changed _how_ the package is built (for example, pinning a dependency) without modifying the source (which would imply a _package_ version bump).

I'll wait to see if Travis is successful and then merge.

jaimergp:
You should be able to merge when tests pass, @hannahbrucemacdonald. I won't monitor this further tonight, it's kind of late here :D

PR #1026, opened 2020-03-12
[perses] 0.5.1 build 3 : Stay away from latest broken numpy 1.18.1

jchodera:


jchodera:
Conda-forge has fixed this upstream.

PR #1037, opened 2020-06-01
[openforcefields] 1.2.0

j-wags:


jchodera:
@j-wags: Any chance this has a mechanism for programmatically identifying which force fields are installed by this repo? I'd love to not have to manually update https://github.com/openmm/openmmforcefields every time

j-wags:
Not yet -- This is tracked in https://github.com/openforcefield/openforcefield/issues/477

If you need a short-term workaround, the following test function should be able to help: https://github.com/openforcefield/openforcefields/blob/master/openforcefields/tests/test_openforcefields.py#L19-L26

PR #1039, opened 2020-06-04
Perses 0.7.0

hannahbrucemacdonald:


perses/meta.yaml
```diff
@@ -22,7 +22,8 @@ requirements:
     - numpy >=1.14,<1.18 # [py36]
     - scipy
     - pymbar
-    - openmm >=7.3.1
+    - openmm ==7.4.2
```
jchodera, line 25: This should be `>=7.4.2` or else it will break other people's environments.

jchodera: I will fix and merge.

PR #1043, opened 2020-06-09
Update numpy defaults

mattwthompson:


mattwthompson:
Hmm, the Python 3.6 build now may work (it did not raise the previous error) but the Python 3.7 build is now crashing. Will investigate more tomorrow.

PR #1052, opened 2020-06-24
[openmmforcefields] 0.7.4

jchodera:


j-wags:
No luck on this so far. We have manually uploaded a locally-built 0.7.4 package to omnia, which seems to be working. So I'm going to stop trying to make this work.

PR #1053, opened 2020-06-28
[openff-evaluator] 0.1.1 build 0 (label: main)

SimonBoothroyd:


j-wags:
I have built this locally and uploaded to omnia. The new package passes tests on linux and OSX. It can be seen here: https://anaconda.org/omnia/openff-evaluator/files

PR #1054, opened 2020-06-29
[perses] version 0.7.1

hannahbrucemacdonald:


j-wags:
@hannahbrucemacdonald My attempt to locally package `perses` is unable to run tests because it doesn't pull down `nose`. Do you mind if I add that to the recipe?

jchodera:
Please do!

hannahbrucemacdonald:
My bad! Sorry!

Thanks @j-wags !

j-wags:
Ok, I think I've got this worked out. Since I built this locally, I don't want the automated build system to see it, so I'm moving `perses/` to a new `local_builds/` subdirectory until we get the automated builds back.

Tests are passing for me on mac and linux:
```
git clone git@github.com:choderalab/perses.git

cd perses

git checkout 0.7.1

conda create --yes -n temp_perses -c conda-forge -c omnia/label/rc -c omnia -c openeye perses 

conda activate temp_perses

GITHUB_ACTIONS=true nosetests perses --verbosity=3 --with-timer -a '!advanced'

```

If there are no objections in the next two hours, I'll move this package to the `main` label and merge this PR.

jchodera:
@j-wags : Thanks so much for helping with this! 

I think there's a fatal flaw here: We can't use `noarch` since the `setup.py` contains an entry-point:
```

      entry_points={
       'openmm.forcefielddir' : ['perses=perses:get_datadir'],
       'console_scripts' : ['perses-relative = perses.app.setup_relative_calculation:run','perses-fah = perses.app.fah_generator:run']
      },
```
When you package with python 3.6, we end up with the entry point command-line executable that _only_ works if your environment is python 3.6:
```
#!/home/chodera/miniconda/envs/perses/bin/python3.6
# -*- coding: utf-8 -*-
import re
import sys
from perses.app.fah_generator import run
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(run())
```
So it's not really noarch as a result.

PR #1056, opened 2020-07-03
[qcsubmit] 0.0.0

j-wags:
To install:
```
conda create -n temp_qcsubmit -c conda-forge -c omnia/label/rc -c omnia -c openeye qcsubmit
conda activate temp_qcsubmit
pip install basis_set_exchange
```

Works for me on mac, fails with a version/git error on Linux.

@jthorton, could you try out the installation steps above and let me know if this works to your satisfaction (especially on Linux, if you have a machine available)?

### DIFF
--- a/offline-builds/qcsubmit/README.md
+++ b/offline-builds/qcsubmit/README.md
@@ -1,0 +1,17 @@
+This is a recipe for building the current development package into a conda
+binary.
+
+The installation on travis-ci is done by building the conda package, installing
+it, running the tests, and then if successful pushing the package to binstar
+(and the docs to AWS S3). The binstar auth token is an encrypted environment
+variable generated using:
+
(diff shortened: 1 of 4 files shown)


jthorton:
@j-wags Thanks for putting this together! it's working on mac and Linux for me with all tests passing. Is there a way to get the `basis_set_exchange` package included in the install as well?

j-wags:
I've tried making a few changes to bring this in line with how Perses gets built. I also tried a few ways to pip install `basis-set-exchange` (such as `${PYTHON} -m pip install basis-set-exchange` and vanilla `pip install basis-set-exchange`, with and without quotes around the package name). It keeps failing with a "package not found" error. 

> ERROR: Could not find a version that satisfies the requirement basis-set-exchange (from versions: none)
> ERROR: No matching distribution found for basis-set-exchange

To reproduce, make an environment with `conda-build` and run the `build_local.sh` script in qcsubmit.

PR #1058, opened 2020-07-05
[openmmtools] 0.20.0

jchodera:


jchodera:
This has now been built and pushed for `linux` and `win` for python 3.6 and 3.7.

PR #1060, opened 2020-07-07
[Perses] 0.7.1 build 6

j-wags:
Trying to see if we can get a noarch build that doesn't have the problems of builds 0 and 1.

Removed build.sh, moved pip install to meta.yaml, included entry points in meta.yaml

To test:
`conda create --yes -n temp_perses6 -c conda-forge -c omnia/label/rc -c omnia -c openeye perses`

### DIFF
--- a/offline-builds/perses/meta.yaml
+++ b/offline-builds/perses/meta.yaml
@@ -13,7 +13,7 @@ build:
     - perses-relative = perses.app.setup_relative_calculation:run
     - perses-fah = perses.app.fah_generator:run
   script: 
-    - ${PYTHON}  -m pip install .  preserve_egg_dir: True
+    - ${PYTHON}  -m pip install .
 requirements:
   build:
     - python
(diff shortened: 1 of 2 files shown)


offline-builds/perses/meta.yaml
```diff
@@ -5,9 +5,15 @@ source:
...
+  noarch: python
   preserve_egg_dir: True
-  number: 4
+  number: 6
   skip: True  # [win or py2k]
+  entry_points:
+    - perses-relative = perses.app.setup_relative_calculation:run
+    - perses-fah = perses.app.fah_generator:run
+  script: 
+    - ${PYTHON}  -m pip install .  preserve_egg_dir: True
```
jchodera, line 16: Should `preserve_egg_dir: True` really be on this line?

j-wags: Good catch -- Fixed. 

I had to reconstruct this file after I accidentally committed to master instead of a branch. The real package was build from a good `meta.yaml` -- This was just a copy-paste error.

j-wags:
Moved to `main` label, merging!

PR #1065, opened 2020-08-28
Perses 0.8.1 release

hannahbrucemacdonald:
Sorry Jeff

### DIFF
--- a/offline-builds/perses/meta.yaml
+++ b/offline-builds/perses/meta.yaml
@@ -1,9 +1,9 @@
 package:
   name: perses
-  version: 0.8.0
+  version: 0.8.1
 source:
   git_url: https://github.com/choderalab/perses.git
-  git_tag: 0.8.0
+  git_tag: 0.8.1
 build:
   noarch: python
   preserve_egg_dir: True


j-wags:
No need to apologize! I'll see if I can get this out before the weekend.

j-wags:
0.8.1 package is built and uploaded to omnia/label/rc. All tests pass on mac with py36/37. I'm unable to test on linux since UCI's HPC is having file system issues.

In a few minutes I'm heading out of internet range for the weekend. I think it might be reasonable to just switch this onto the `main` label and be ready to roll back in case of emergency, since I almost never see OS-specific problems in this mature a package. I'll leave it up to @jchodera if you'd like it before the weekend.

If you have a linux box and are interested in testing before I return on Monday, my dance routine is
```
conda create -n temp_perses081_36 -c conda-forge -c omnia/label/rc -c omnia -c openeye perses python=3.6
conda activate temp_perses081_36
GITHUB_ACTIONS=true nosetests perses --verbosity=3 --with-timer -a '!advanced'
```

jchodera:
Released to `main`! Thanks again @j-wags!

j-wags:
Excellent -- Just cleaning up this PR/branch for recordkeeping.

PR #1066, opened 2020-09-10
[openforcefields] 1.2.1

j-wags:


j-wags:
Package uploaded and release passes tests. Moved to omnia main label.

PR #1067, opened 2020-09-19
[fragmenter] 0.0.4 build 1

j-wags:
Fragmenter is incompatible with OE2020, per https://github.com/openforcefield/fragmenter/issues/61

Putting this up as a reminder -- Will build package and test next week.


j-wags:
Built, uploaded, and tested. Moving to main :-)

PR #1069, opened 2020-09-30
[fragmenter] 0.0.7

j-wags:
@jthorton just pointed out that 0.0.4 is no longer the latest version of fragmenter, and actually has a conflict with the latest CMILES currently on conda forge. He can fix it by installing the latest fragmenter from master. So I'm updating the version of this package to the newest tag (0.0.7)


j-wags:
I haven't heard any complaints about this, so I'll push the package to the main label. We can make a new build if there are dependency resolution issues.

PR #1079, opened 2020-12-18
Update smirnoff99frosst: noarch, entry_points test, no cython

mattwthompson:
I was beaten to the punch for the meat of this, but I'm opening this anyway to see if `run_test.py` works out of the box here (it did locally :tm: ) - if yes, will add to the `conda-forge` recipes

### DIFF
--- a/smirnoff99frosst/build.sh
+++ b/smirnoff99frosst/build.sh
@@ -1,4 +1,0 @@
-#!/bin/bash
-
-$PYTHON setup.py clean
-$PYTHON setup.py install
--- a/smirnoff99frosst/meta.yaml
+++ b/smirnoff99frosst/meta.yaml
@@ -7,11 +7,13 @@ source:
   git_tag: 1.1.0
 
 build:
-  number: 1
-  #noarch: python
+  number: 2
+  noarch: python
   entry_points:
     - smirnoff99frosst = smirnoff99frosst.smirnoff99frosst:main
-  skip: True  # [win32 or (win and py2k)]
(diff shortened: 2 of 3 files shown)


j-wags:
Ah, interesting. I hadn't thought of testing at this stage, but that could be cool. For my information -- Does conda-forge recommend doing build-time tests?

mattwthompson:
It's uncommon, so I don't think it's considered best practices, but it's definitely allowed. `conda-build` lets you drop in a script, in addition to any set of commands you'd like. You probably already know about [this](https://github.com/conda-forge/mdtraj-feedstock/blob/a6707ed05d66d8b27d5011ab15e2457057187ae0/recipe/meta.yaml#L41-L47) pattern, and there's nothing strictly stopping one from running the [full test suite](https://github.com/mosdef-hub/mbuild/blob/240b30c782a7fa7a0c3b5c5bc5ed7941ad90d889/devtools/conda-recipe/meta.yaml#L41-L44). However, I think it's pretty easy to argue that import tests alone will cover virtually anything broken at build time, assuming the package was stable when the release was created. In this case, I want to check if the entry points were wired up correctly, which doesn't fit into the scope of unit tests (debatable?)  and wouldn't be caught with an import test, so that's why I used a build-time [test script](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-test-script). There may be a better way to catch this, but this is already a huge improvement over the very, very downstream nature of how I caught this today.

